PR #6, opened 2022-07-27
Add spiral route planning order

tomasloksa:
+ Added spiral route planning order with variable spiral size. 
+ Updated docs for route planning

If you are okay with the proposed changes, please let me know and i will also add tests.
 
I don't have much experience with c++ projects, so any comments are more than welcome :smile:


Gonzalo-Mier:
Thank you for your contribution.  I don't really understand the pattern. Could you add examples with the order of 6 swaths and 7 swaths. Great if it's explained in the docs. For me the contribution is great.

Don't forget to add yourself to the Contributors.md and AUTHORS file

tomasloksa:
I tried to add a better explanation to the docs. Hope it makes everything clear now. But  i think and image would make everything even more clear.
Should i add numbers to the swaths on the image, or maybe even Dubins paths on the headland could be enough...

Gonzalo-Mier:
I think what you have now is enough. Create some tests and whenever you are done I will accept the PR :) Thank you for your contribution.

tomasloksa:
I created tests for even/odd spiral sizes. 
I couldn't quite make sense of the computeCost function and couldn't figure out an equation, that would satisfy the condition, so i left that out.

PR #7, opened 2022-08-03
Python interface

jfulem:
I added the spiral order for the swaths. Also, I think it would be good to have one more constructor of the `SpiralOrder` with only one argument being the size of the spiral. Then similarly to others, the `setSwaths(F2CSwaths& swaths)` method can be used.


Gonzalo-Mier:
Hi @jfulem,
First, thank you for your contribution. Unfortunately, I would split it in two different PR. First, the new constructor of spiral that should go to the `main` branch, and second, the `python_interface` that merge all the changes to the second branch. If we accept this PR, the constructor will be only in the `python_interface` branch until the `python_interface` is merged to `main`. Do you agree about it? Could you split it in two PR, please?

Also, remember to add yourself to the `Authors` and `Contributors` files. Thank you again.

jfulem:
All right. I'll do it.

Gonzalo-Mier:
@jfulem Please, tell me when the PR is ready to review.

jfulem:
I think it is.

Gonzalo-Mier (approved):
Thank you for your contibution!

PR #15, opened 2022-08-29
Custom order swaths

jfulem:
The swaths order is given by the custom order vector.


Gonzalo-Mier:
Hi @jfulem, thank you again for your contributions. I found your PR a great start, but I would like to ask you some improvements.

1. We should suppose we don't know how many swaths a field contains. What happens if the field contains fewer swaths than the pattern? And if the field has three times more swaths than the pattern provided? Have you tried if both cases it works? It would be nice to see some tests covering those cases.
2. Could you also provide some boundary checks with the provided order? Throw an exception if `f2c::rp::CustomOrder swath_sorter(swaths, {10,2});` because the order doesn't have 0 as the lowest value and `order.size() - 1` as the maximum. Also check if any value is repeated.
3. Arguments on functions should be passed by const reference by default (if it is not a basic type like an int or a double). If not, it should have a strong reason to do it in that way. For example, If I initialize a CustomOrder using `f2c::rp::CustomOrder(F2CSwaths& swaths, std::vector<size_t> order)`, the function caller would copy the vector `order`, and then copy it again when this function calls to `void CustomOrder::set_custom_order(std::vector<size_t> order)`, and finally inside `set_custom_order` the vector `order`  is copied again. Instead of copying the vector once, it's copied 3 times. Not a big issue as those vectors usually are small, but it's an easy improvement.
4. Following the previous comment, `void CustomOrder::sort_swaths(F2CSwaths& swaths, std::vector<size_t>& order)` pass `order` by reference, but it is not constant. That makes `order` modifiable. Unfortunately, when `void CustomOrder::sort_swaths(F2CSwaths& swaths, std::vector<size_t>& order)` is called, `this->custom_order` is also modified. If other swaths are set to the sorter algorithm, and then sorted, the result is undefined.
5. Convention for function names are Uppercase the first letter of the second/third/fourth... word without underscores as separation. So `void sort_swaths(F2CSwaths& swaths, std::vector<size_t>& order)` should be `void sortSwaths(F2CSwaths& swaths, std::vector<size_t>& order)`. Same with `void set_custom_order(std::vector<size_t> order)` to `void setCustomOrder(std::vector<size_t> order)` (I should write a style guide to avoid this kind of comments for the future. It's my fault).
6. The solution of `void CustomOrder::sort_swaths(F2CSwaths& swaths, std::vector<size_t>& order)` seems very clever, but I wasn't able to fully understand it. How did you choose this algorithm? Could I propose instead doing something like (I haven't tested the code):
  ```
  // Improvements number 1, 2 and 5 remains unsolved with this solution. 
  void CustomOrder::sort_swaths(F2CSwaths& swaths, const std::vector<size_t>& order) {
    F2CSwaths sorted_swaths (swaths.size());
    for (size_t i = 0; i < order.size(); ++i) {
      sorted_swaths[i] = swaths[order[i]];
    }
    swaths = std::move(sorted_swaths);
  }
  ```

Sorry for so many comments. I believe the code is pretty nice and only small changes are needed, but each comment are really verbose. Thanks again

Gonzalo-Mier:
Hi @jfulem, please, always tell me when you are done with changes to review it. I've realized that documentation is missing. Do you mind adding it? Thank you :)

tests/python/route_planning/test_custom_order.py
```diff
@@ -0,0 +1,48 @@
...
+        f2c.Swath(
+          f2c.LineString(
+            f2c.VectorPoint([
+              f2c.Point(0, i),
+              f2c.Point(1, i)])), 
+          1, 
+          i
+        )
+    )
+  swath_order = f2c.RP_CustomOrder(swaths, f2c.VecSize(custom_order))
```
Gonzalo-Mier, line 29: pytest error. VecSize doesn't exist. Do you mean VectorSize?

tests/python/route_planning/test_custom_order.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.mark.parametrize('n,custom_order,expected', PARAMETRIZED_INPUTS_WRONG)
+def test_fields2cover_wrong_custom_order(n, custom_order, expected):
+  swaths = f2c.Swaths()
+  for i in range(n):
+    swaths.push_back(
+        f2c.Swath(
+            f2c.LineString(
+                f2c.VectorPoint([f2c.Point(0, i),
+                                 f2c.Point(1, i)])), 1, i))
+  swath_order = f2c.RP_CustomOrder(swaths, f2c.VecSize(custom_order))
```
Gonzalo-Mier, line 44: pytest error. VecSize doesn't exist. Do you mean VectorSize?

Gonzalo-Mier (changes requested):
Errors on pytest

tests/python/route_planning/test_custom_order.py
jfulem: Yes, I just did not have time to change this.

tests/python/route_planning/test_custom_order.py
jfulem: Yes, similar as above

src/fields2cover/route_planning/custom_order.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  if (swaths_.size() != custom_order.size()) {
+    throw std::length_error("Lengths of the order vector and swaths must be the same.");
+  }
+  auto max = *std::max_element(custom_order.begin(), custom_order.end());
+  if (max > swaths_.size() - 1){
+    throw std::invalid_argument("Custom order element is out of the swath range [" + std::to_string(max) + "]");
+  }
+  //
+  if (diff.size() > 0) {
```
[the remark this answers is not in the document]
jfulem: I think it was overcomplicated - I tried to report what elements are wrong. I think you are right.

jfulem:
@Gonzalo-Mier The pytest should be fixed as well as the rest.

jfulem:
You are welcome. I tried to compile everything on MacOS. I'm pretty close, but I'm not sure if it is worth. Thanks.

PR #19, opened 2022-09-14
remove redundant clone that caused memory leak

philipogorman:
We found a memory leak in `Point`, there is a redundant clone that does not cleaned up. The fix is simple, just remove the extra clone.
Below is a screen shot of the valgrind report on the unit test. 

image

### DIFF
--- a/src/fields2cover/types/Point.cpp
+++ b/src/fields2cover/types/Point.cpp
@@ -61,7 +61,7 @@ Point Point::operator*(double b) const {
   return Point(data->getX() * b, data->getY() * b, data->getZ() * b);
 }
 
-Point Point::clone() const {return Point(data->clone());}
+Point Point::clone() const {return Point(data);}
 double Point::getX() const {return data->getX();}
 double Point::getY() const {return data->getY();}
 double Point::getZ() const {return data->getZ();}


philipogorman:
made an issue https://github.com/Fields2Cover/Fields2Cover/issues/20

Gonzalo-Mier:
Hello @philipogorman, 
Thanks for your contribution. I had problems running valgrind because of python on matplotlib-cpp. How did you solve it? If it is not a huge problem, could you make a small bash script to run valgrind on the library?

Running `make check` should have given you an error. I have pushed a new test to check if the clone of a Point is correct.
Could you try with this line and valgrind instead?
```
Point Point::clone() const {return Point(data->getX(), data->getY(), data->getZ());}
```

Thank you really much.
Best,
Gonzalo

Gonzalo-Mier:
Thanks for your contribution @philipogorman!! 
I'll try CLion soon. If needed, I'll contact you again. Thank you again!

philipogorman:
@Gonzalo-Mier please do contact me if you run into questions when using clion - I'd be happy to help.
We have been testing this repo and are very impressed with the quality of the code. Thank you!

Issue #14, opened 2022-08-24
Random class is based on rand_r(). Better if rand_r() is replaced by the <random> library


Gonzalo-Mier:
This is a good issue to get used to the repository. `rand_r` is obsolete by POSIX.1-2008, which means in the future it may not work anymore. 

To solve this issue, the `f2c::Random` class (on files `include/fields2cover/utils/random.h` and `src/fields2cover/utils/random.cpp`) should be changed to use functions in the `<random>` lib instead of rand_r

Gonzalo-Mier:
Cool video about this issue: https://www.youtube.com/watch?v=LDPMpc-ENqY

mdabek:
You can assign this issue to me.

PR #30, opened 2022-10-05
Random generators based on the STL random.

mdabek:
I've also added new exponential random generator, which is compatible with the exponential distribution. It generates random numbers from range [0; +inf) and takes single parameter, lambda -the  rate parameter.
The unit test bases on the distribution property, that the: E[X] = 1/lambda.
This PR is linked to the #14

### DIFF
--- a/include/fields2cover/utils/random.h
+++ b/include/fields2cover/utils/random.h
@@ -57,8 +57,8 @@ class Random {
   /// probability.
   /// Lower values are more probable than higher ones.
   /// @param lambda is the rate parameter
-  /// @return Double \f$\in [min, max]\f$
-  double getRandomExp(double lambda);
+  /// @return Double > 0 
+  double getRandomExpDist(double lambda);
 
   /// Return a random angle in radians.
   /// @return random angle \f$ \in [0, 2\pi) \f$
--- a/include/fields2cover/utils/random.h
+++ b/include/fields2cover/utils/random.h
@@ -10,6 +10,7 @@
 
 #include <time.h>
 #include <limits>
+#include <random>
 #include <boost/math/constants/constants.hpp>
 #include "fields2cover/types/Point.h"
 #include "fields2cover/types/LineString.h"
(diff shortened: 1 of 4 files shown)


include/fields2cover/utils/random.h
```diff
@@ -52,6 +53,13 @@ class Random {
   /// @return Double \f$\in [min, max]\f$
   double getRandomExp(double min, double max);
 
+  /// Get a random double in a range [0;+inf) with exponential
+  /// probability.
+  /// Lower values are more probable than higher ones.
+  /// @param lambda is the rate parameter
+  /// @return Double \f$\in [min, max]\f$
```
Gonzalo-Mier, line 60: This comment is not true

include/fields2cover/utils/random.h
```diff
@@ -52,6 +53,13 @@ class Random {
...
+  double getRandomExp(double lambda);
```
Gonzalo-Mier, line 61: getRandomExp do different things when has only one parameter than when it has two. One could expect the single parameter version to return a value between (0, lambda) which is not the case. Could you put another name to this function? Like getRandomExpDist(double lambda)

Gonzalo-Mier (changes requested):

Gonzalo-Mier:
Thanks for your contribution!!! Could you please add the same test you add it, but to the python tests? So both versions benefits from this new function 😁 Really good job

include/fields2cover/utils/random.h
Gonzalo-Mier: The @return part

include/fields2cover/utils/random.h
mdabek: Yes. I will change the method name.

include/fields2cover/utils/random.h
mdabek: Right. Will fix this.

Gonzalo-Mier (approved):

Gonzalo-Mier:
Thank you again!

PR #37, opened 2022-11-01
Add support for the ROS build system

Timple:
Note that this does not make this library a ROS package. It will still be standalone.

However for people within a ROS ecosystem, this would greatly improve the build experience (and drop a `sudo make install` command).

Can be verified with this dockerfile (which might be added to CI):

```Dockerfile
FROM ros:noetic
RUN mkdir -p /ws/src
WORKDIR /ws
RUN apt-get update -qq && apt-get install -qqy python3-catkin-tools python3-osrf-pycommon git software-properties-common

RUN add-apt-repository -y ppa:ubuntugis/ppa && apt-get update -qq

RUN git clone https://github.com/nobleo/Fields2Cover src/fields2cover --branch compile-with-ros
RUN git clone https://github.com/Fields2Cover/fields2cover_ros src/fields2cover_ros
RUN rosdep install -r --ignore-src -y --from-paths .

RUN bash -c 'source /opt/ros/*/setup.bash && catkin build'
RUN bash -c 'source /opt/ros/*/setup.bash && catkin run_tests'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,14 +102,14 @@ foreach(p LIB BIN INCLUDE CMAKE)
   endif()
 endforeach()
 
-file(GLOB_RECURSE fields2cover_src 
+file(GLOB_RECURSE fields2cover_src
     "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp"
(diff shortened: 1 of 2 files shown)


Gonzalo-Mier:
Awesome!!! Thank you for contributing. I believe this is actually a great addition

PR #93, opened 2023-10-01
Feature/swath discretization

Marnonel6:
# Description:

This PR introduces a new feature in our Path Planning module which discretizes swath sections of the path. By providing a discretization step and the robot's swath speed, the function returns a new path with the swath sections discretized.

# Main Changes:

Swath Discretization Function:

- Added a new method `Path::discretize_swath(double step_size, double swath_speed)` that takes in a `step_size` and `swath_speed` and returns a new path with discretized swath sections.
- The method works by iterating through the path points and discretizing swath points using the given step_size. The velocity of the swath is set based on the provided swath_speed.

## Example Script:

- An example script `discretize_swath_path.cpp` demonstrates the use of the `discretize_swath` function.
- This script defines a field, a robot, and generates swath connections using Dubins curves. After which, it uses the new `discretize_swath` function to generate a discretized path and visualize it using our visualization module.

## Tests:
- Different swaft, headland and path planning parameters were tested.
- The discretized path was visualized and inspected for correct discretization.
- Graphs were also plotted of the coordinates and the angle to ensure correctness.

## Screenshots & Files
image
[discretized_swath_path.csv](https://github.com/Fields2Cover/Fields2Cover/files/12778375/discretized_swath_path.csv)


src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+    for (size_t i = 0; i < this->size(); i++)
+    {
+        // Check if the current point is a SWATH point
+        if (this->states.at(i).type == f2c::types::PathSectionType::SWATH)
+        {
+            // We know that swaths are represented by only two point so the next point will
+            // be the end of the swath
+            // Save start and end swath points to variables
+            Point start_point = this->states.at(i).point;
+            Point end_point = this->states.at(i+1).point;
```
Gonzalo-Mier, line 137: if (i == this->size() - 1) then this->states.at(i+1) gets out of the boundaries

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+
+            // We can now discretize the swath
+            // Calculate the distance between the start and end point
+            double distance = start_point.Distance(end_point);
+            // Calculate the number of points needed to discretize the swath
+            int num_points = floor(distance / step_size);
```
Gonzalo-Mier, line 143: with floor, distances between points will be bigger than step_size. Why don't you use ceil?

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+            for (int j = 0; j < num_points; j++)
+            {
+                x_values.push_back(start_point.getX() + (j * (end_point.getX() - start_point.getX()) / num_points));
+                y_values.push_back(start_point.getY() + (j * (end_point.getY() - start_point.getY()) / num_points));
+                z_values.push_back(start_point.getZ() + (j * (end_point.getZ() - start_point.getZ()) / num_points));
+            }
+            // Add last point to list
+            x_values.push_back(end_point.getX());
+            y_values.push_back(end_point.getY());
+            z_values.push_back(end_point.getZ());
```
Gonzalo-Mier, line 158: if you make the previous loop instead of (for (int j = 0; j < num_points; j++)) --> (for (int j = 0; j <= num_points; j++)), these three lines are redundant.

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+
+            // Iterate over each pair of coordinates and add them as states into our new Path object
+            for (int j = 0; j < num_points; j++)
+            {
+                // Create a new PathState object
+                PathState state;
+                // Set the point to the current x and y values
+                state.point = Point(x_values.at(j), y_values.at(j), z_values.at(j));
```
Gonzalo-Mier, line 166: Why x_values are not computed here directly instead of creating a vector for it?

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+                // Set the angle to the angle between the current point and the next point
+                state.angle = atan2(y_values.at(j+1) - y_values.at(j), x_values.at(j+1) - x_values.at(j));
```
Gonzalo-Mier, line 168: Is this angle different from the initial one from the start point to the end point?

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+                // Ensure the angle is in the range [0, 2*PI]
+                if (state.angle < 0.0){
+                    state.angle = state.angle + 2 * M_PI;
+                }
+                // Set the velocity to the swath/cruise speed of the robot
+                state.velocity = swath_speed;
```
Gonzalo-Mier, line 174: Why not use this->states.at(i).velocity instead of modifying it?

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+                // Set the duration to the distance between the current point and the next point
+                state.duration = sqrt(pow(x_values.at(j+1) - x_values.at(j), 2) + pow(y_values.at(j+1) - y_values.at(j), 2));
```
Gonzalo-Mier, line 176: duration is distance (you previously computed it) / velocity

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+                state.dir = f2c::types::PathDirection::FORWARD;
+                state.type = f2c::types::PathSectionType::SWATH;
+                new_path.states.push_back(state);
+            }
+            i++; // Increment i with one to move past end_point of current swath
```
Gonzalo-Mier, line 181: What happens if the swath has more than two points?

src/fields2cover/types/Path.cpp
```diff
@@ -114,6 +114,81 @@ std::string Path::serializePath(size_t digit_precision) const {
...
+            // Add last point to list
+            x_values.push_back(end_point.getX());
+            y_values.push_back(end_point.getY());
+            z_values.push_back(end_point.getZ());
+
+            // Iterate over each pair of coordinates and add them as states into our new Path object
+            for (int j = 0; j < num_points; j++)
+            {
+                // Create a new PathState object
+                PathState state;
```
Gonzalo-Mier, line 164: with "PathState state = this->states.at(i);" you can avoid lines 168-174 and 177-178

tutorials/discretize_swath_path.cpp
Gonzalo-Mier, line 1: This should be a test, not a tutorial

Gonzalo-Mier (changes requested):
Hi @Marnonel6, cool PR!! Thank you for your contribution. I have left a few comments to solve a few issues I've seen. Thank you again

Marnonel6:
I'll get on those comments!

src/fields2cover/types/Path.cpp
Marnonel6: Does this mean that sometimes the very last element in the path can be a single swath element? As far as I've seen swaths always have two point's a start and end when using the straight line swaths. If this is not the case please let me know what all the possible cases are for swaths in the Path objects.

src/fields2cover/types/Path.cpp
Marnonel6: # Just a simple example to make sure we are on the same page:
- step_size = 3[m]
- swath_distance = 10[m]
- swath_start = 0[m]
- swath_end = 10[m]
## Floor
- num_points = floor(10/3) = floor(3.33) = 3 -> This will mean the new swath elements [0, 3, 6, 9, 10] thus the last step size will be the residual which is 1[m]
## Ceil
- num_points = ceil(10/3) = floor(3.33) = 4 -> This will mean the new swath elements [0, 3, 6, 9, 12] thus the last element is now 12[m] instead of 10[m]
## Other solution
Instead of using exactly the step size provided the closest approximate step size can be calculated to give equal steps between swaths with something like this:
- rounded_number_of_steps = std::round(swath_distance / step_size);

Case where step_size > swath_distance 

               // If rounded number of steps is zero, then the provided step_size is greater than the distance.
               // In this case, we should use the distance itself as the step size, thus no discretization needed.
               if(rounded_number_of_steps == 0) {
                     approx_step_size = distance;
                }

- approx_step_size = distance / rounded_number_of_steps

This will result in equal spaced steps from start to finish, but with a step_size that tries to approximate the user step_size. Let me know what you think.

src/fields2cover/types/Path.cpp
Marnonel6: Fixed, thank you.

src/fields2cover/types/Path.cpp
Marnonel6: Fixed, Thank you!

src/fields2cover/types/Path.cpp
Marnonel6: When using straight swaths no. So I can use the original swath angle at the start point as the angle throughout the swath.

src/fields2cover/types/Path.cpp
Marnonel6: Updated.

src/fields2cover/types/Path.cpp
Marnonel6: You are correct. 

New method of calculating:
state.duration = static_cast<double>((distance/state.velocity)/rounded_number_of_steps);

src/fields2cover/types/Path.cpp
Marnonel6: I was not aware that it can have more than two points apologies. As far as I knew swaths where straight lines with a start and end point defining it. Please let me know what other swaths configurations might exist.

tutorials/discretize_swath_path.cpp
Marnonel6: Removed, thanks.

Marnonel6:
Updated some of the mentioned comments. Let me know what you think and what should still change @Gonzalo-Mier

src/fields2cover/types/Path.cpp
Marnonel6: I went with the `Other solution`

src/fields2cover/types/Path.cpp
Gonzalo-Mier: As mentioned above, swaths can have more than two points

Gonzalo-Mier (changes requested):
Great! Could you add some test to [](https://github.com/Fields2Cover/Fields2Cover/blob/main/tests/cpp/types/Path_test.cpp) to check that it works as expected in any future version? Thank you

Marnonel6:
Okay I've updated your latest comments and added a bunch of tests. Let me know what you think.

Gonzalo-Mier:
Thank your really much. Just for the record, I made a few changes. I changed the style of the code to be in the same style as the rest of the repo (2 spaces instead of 4 and { in the same line as for() and if()). Moreover, I moved the test from the path_planning to the Path_test.
Great job!!

Marnonel6:
My pleasure. Thank you for the merge!

PR #102, opened 2023-12-14
Fail silently for cartesian systems

SteveMacenski:
Currently the parser has very strict interpretations of `srsName`. This adds a more relaxed option for parsing files that have empty or non-GPS frames (e.g. `map`). By default, nothing changes, you have to explicitly enable this flag to allow for silent failing of coord parsing with `createSptRef` messing with references.


Gonzalo-Mier:
Thank you!! Nice PR.

PR #103, opened 2023-12-17
swath and swaths transform with reference points

SteveMacenski:
Does the same as the `path` for the swath data, I believe. I use the other transform function you have instead of manually typing it all out like you do in `transformPathWithFieldRef`. I believe these are identical operations however. 

I also added some coord reference `moveTo` to swaths like are in `path` for the non-GPS case when I need to add a reference point back to some data

### DIFF
--- a/include/fields2cover/types/Swath.h
+++ b/include/fields2cover/types/Swath.h
@@ -91,6 +91,9 @@ struct Swath {
   /// less than pi
   void targetOppositeDirAs(const Swath& s);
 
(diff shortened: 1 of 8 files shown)


Gonzalo-Mier:
Thank you for the PR. Could you add a few tests for the new functions? Thank you again!

SteveMacenski:
Done

Gonzalo-Mier:
Cool! Thank you

PR #125, opened 2024-04-24
Fix cmake install locations for ROS2/colcon

aosmw:
* Work around non-standard (for ROS2) location of toplevel include file fields2cover.h that refers to sub includes via "relative paths" mechanism by also including $<INSTALL_INTERFACE:include/fields2cover> in target_include_directores(Fields2Cover ...

* Install cmake support files in ROS2 standard/convention directory share/fields2cover/cmake


Gonzalo-Mier:
Thank you really much @aosmw !! I didn't know how to solve this and it was driving me crazy

PR #126, opened 2024-04-29
Feature/mw/deps fixes

aosmw:
Use ortools_vendor if available, and find_package(TinyXML2 ...) better.


Gonzalo-Mier:
Thank you again! Let's try it :)

PR #129, opened 2024-04-30
add redirect flag

ar0ser0:
Adds a flag to change the direction (or not) of the given swaths to `genRoute()`. This is useful if you care about the swath pattern/direction. This will allow users to use `SingleCellSwathsOrderBase` class with `genRoute()`. This does not enforce order, but it does enforce the pattern (direction).

### DIFF
--- a/include/fields2cover/route_planning/route_planner_base.h
+++ b/include/fields2cover/route_planning/route_planner_base.h
@@ -32,8 +32,8 @@ class RoutePlannerBase {
   /// @param d_tol Tolerance distance to consider if two points are the same.
   /// @return Route that covers all the swaths
   virtual F2CRoute genRoute(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths = true,
-      bool show_log = false, double d_tol = 1e-4);
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
+      bool show_log = false, double d_tol = 1e-4, bool redirect_swaths = true);
 
   /// Set the start and the end of the route.
   void setStartAndEndPoint(const F2CPoint& p);
@@ -56,9 +56,9 @@ class RoutePlannerBase {
   ///          between two nodes.
   /// @param d_tol Tolerance distance to consider if two points are the same.
   virtual F2CGraph2D createCoverageGraph(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths,
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
       F2CGraph2D& shortest_graph,
-      double d_tol) const;
+      double d_tol, bool redirect_swaths = true) const;
 
 
  protected:
--- a/include/fields2cover/route_planning/route_planner_base.h
+++ b/include/fields2cover/route_planning/route_planner_base.h
@@ -32,7 +32,7 @@ class RoutePlannerBase {
   /// @param d_tol Tolerance distance to consider if two points are the same.
   /// @return Route that covers all the swaths
   virtual F2CRoute genRoute(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths = true,
       bool show_log = false, double d_tol = 1e-4);
 
   /// Set the start and the end of the route.
@@ -56,7 +56,7 @@ class RoutePlannerBase {
   ///          between two nodes.
   /// @param d_tol Tolerance distance to consider if two points are the same.
   virtual F2CGraph2D createCoverageGraph(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths,
       F2CGraph2D& shortest_graph,
       double d_tol) const;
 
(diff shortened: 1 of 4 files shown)


include/fields2cover/route_planning/route_planner_base.h
```diff
@@ -32,7 +32,7 @@ class RoutePlannerBase {
   /// @param d_tol Tolerance distance to consider if two points are the same.
   /// @return Route that covers all the swaths
   virtual F2CRoute genRoute(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths = true,
```
Gonzalo-Mier, line 35: A user that was running `genRoute(cells, swaths, false)` will result in a different behavior now. Instead of breaking the compatibility with previous actual version, create a new function.

include/fields2cover/route_planning/route_planner_base.h
```diff
@@ -56,7 +56,7 @@ class RoutePlannerBase {
   ///          between two nodes.
   /// @param d_tol Tolerance distance to consider if two points are the same.
   virtual F2CGraph2D createCoverageGraph(
-      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
+      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells, bool redirect_swaths,
```
Gonzalo-Mier, line 59: This change will prevent compilation of previous code. Add the bool parameter at the end and add a default value that maintains the same behavior as now.

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -107,10 +107,12 @@ F2CGraph2D RoutePlannerBase::createCoverageGraph(
         for (const auto& s2 : swaths2) {
           auto s2_s = s2.startPoint();
           auto s2_e = s2.endPoint();
-          g.addEdge(s1_s, s2_s, shortest_graph);
+          if (redirect_swaths) {
+            g.addEdge(s1_s, s2_s, shortest_graph);
+            g.addEdge(s1_e, s2_e, shortest_graph);
+          }
           g.addEdge(s1_s, s2_e, shortest_graph);
```
Gonzalo-Mier, line 114: I think connections are not properly made in this case. Be aware that `addEdge(a, b)` is bidirectional, so you are connection `a->b` and `b->a`. When reordering is not allowed, you only need `addDirectedEdge(s1_e, s2_s, shortest_graph)`

Gonzalo-Mier (changes requested):
I miss some tests and documentation about those changes,

ar0ser0:
@Gonzalo-Mier I made some changes based on your comments.

src/fields2cover/route_planning/route_planner_base.cpp
ar0ser0: Oh thanks, I did not know that. I made some changes to address this comment. Is there anything else that you know of that can simplify the search space to decrease runtime given that the user knows the direction/pattern of the swaths, but they are ok with having the order rearranged? For example, if they want the arrows  of the swaths to stay fixed in the same direction, but they just want the fastest/shortest route that goes through all the swaths.

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -125,11 +120,23 @@ std::vector<std::pair<F2CPoint, double>> PathPlanning::simplifyConnection(
     }
   }
 
+  if (p1.distance(p2) < 6.0 * safe_dist || vp.size() <2) {
```
Gonzalo-Mier, line 123: Why moving section this down? Early returns are prefered

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -125,11 +120,23 @@ std::vector<std::pair<F2CPoint, double>> PathPlanning::simplifyConnection(
...
+    path.emplace_back(p2, ang2);
+    return path;
+  }
+
   for (int i = 1; i < vp.size() - 1; ++i) {
     double dist_in  = vp[i].distance(vp[i-1]);
     double dist_out  = vp[i].distance(vp[i+1]);
     double d_in  = min(0.5 * dist_in,  safe_dist);
     double d_out  = min(0.5 * dist_out,  safe_dist);
+    // I haven't checked if this is possible, but
```
Gonzalo-Mier, line 133: We were assuming that those points are not duplicated. Maybe a huge assumption. Can you add a little filter?

Gonzalo-Mier (commented):
Can you also add a test to make sure it works as expected?

src/fields2cover/path_planning/path_planning.cpp
ar0ser0: Im not really moving it down. Rather, I'm defining `vp` earlier and using the same early return but using `vp` instead of `mp`.

ar0ser0: I suppose it is going through the for loop to create `vp` each time. We could just add a second check for if `vp.size() < 2` that uses the same return value.

Gonzalo-Mier:
Hi @ar0ser0, can you solve the previous request of adding some tests to check everything works as expected?

ar0ser0:
ok i added a test that follows what was done in the `simple test` of `fields2cover_rp_route_plan_base` but now includes the `redirect flag=false` and checks for directions being the same for each swath

Gonzalo-Mier:
Awesome! Thank you for the PR @ar0ser0

PR #139, opened 2024-05-09
Add find_dependency(ortools REQUIRED) to Fields2CoverConfig.cmake.in

aosmw:
In order for projects like opennav_coverage to use the ortools that would come with fieldstocover we need to add the call to find_dependency to the Fields2CoverConfig.cmake.in

This also requires that the bin directory from the ortools is also "installed"


aosmw:
Note this was part of previous PR related to gnuplot.  It was also reverted when that PR was reverted.

Gonzalo-Mier:
Thank you for the PR

PR #141, opened 2024-05-12
Select ortools release tarball based on CMAKE_SYSTEM_PROCESSOR

aosmw:
This is an attempt to satisfy an linux arm64 build using a prebuilt release tarball.


Gonzalo-Mier:
Cool!

PR #144, opened 2024-05-23
Add swig to package.xml

tonynajjar:
I know swig is only needed for building the python interface but is it so bad to have it as a dependency?


Gonzalo-Mier:
Hi @tonynajjar, thank you for the PR. Merged!

PR #157, opened 2024-06-05
Increase sharpness every iteration to reduce loops

vinnnyr:
I believe that every iteration we want to _increase_ sharpness (diff curvature) to prevent loops rather than _decrease_ it.


Gonzalo-Mier:
Not really. The problem of the **maximum** sharpness is that you cannot (shouldn't) use a value higher than that.

PR #168, opened 2024-07-25
SwathGenerator fix setAllowOverlap

Rick-v-E:
The `setAllowOverlap()` function was not working properly (which was added in #71 ). This PR should fix this.  This option allows to have overlap between the last and one-before-last swath to completely cover the field instead of having the last swath outside the field area.

`setAllowOverlap(False)` (default):

image

`setAllowOverlap(False)`:

image

### DIFF
--- a/src/fields2cover/swath_generator/swath_generator_base.cpp
+++ b/src/fields2cover/swath_generator/swath_generator_base.cpp
@@ -50,16 +50,16 @@ F2CSwaths SwathGeneratorBase::generateSwaths(double angle,
 
   double curve_y {-0.5 * op_width};
   F2CMultiLineString paths;
-  while (field_height > curve_y) {
+  while (field_height > curve_y + (allow_overlap ? 1.5 * op_width : 0.5 * op_width)) {
     curve_y += op_width;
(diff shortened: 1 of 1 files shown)


Gonzalo-Mier:
Hi @Rick-v-E,

Thanks for the PR and sorry for the delay answering. I like the idea of this distribution, and it makes me realize that we have 4 options here. 
For our example, a square field has 37m width and the coverage width is 10m:
1. Without overlapping starting at 0 (5, 15, 25)
2. With overlap at the end (5, 15, 25, 32)
3. Allowing covering outside the field starting at 0 (5, 15, 25, 35)
4. With overlap on the last swath (5, 15, 25, 32); equally distributed overlap (5, 14, 23, 32) 

Right now, we support options 1 and 3, and this PR change 1 for 2. It would be ideal to implement the 4 of them, so the user could pick their preferred behavior.

Rick-v-E:
Hi, 

I think you are right, there are different strategies how to handle with the remaining part of the field. It make sense to let the user pick one of these strategies since it is also depending on the implement they are using. I am not sure where this should be implemented though, and which default to use. We can set an enum in the `Robot` class or in the `SwathGenerator`.

PR #177, opened 2024-08-27
feat: added methods for import

tomschut:
feat: add possibility to set start and end points
feat: add shortest route method
feat: add possibility to import json from string
fix: snake order crashed with a single swath
chore: integrate both docker files with dev container


.devcontainer/Dockerfile
Gonzalo-Mier, line 1: Why are you removing this file? This dockerfile is not the same as the base one, as it is the one used by codespace to run

Dockerfile
```diff
@@ -1,12 +1,26 @@
-FROM osgeo/gdal:ubuntu-full-3.6.3
+FROM osgeo/gdal:ubuntu-full-3.6.3 AS base
+ARG USERNAME
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Run only if username is set
```
Gonzalo-Mier, line 6: Why should we check this?

include/fields2cover/route_planning/route_planner_base.h
```diff
@@ -21,7 +21,7 @@
 namespace f2c::rp {
 
 class RoutePlannerBase {
- public:
```
Gonzalo-Mier, line 24: public, private and protected words should only have one space more than the class name (as before)

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -18,47 +18,66 @@ namespace f2c::rp {
...
-      v_route, swaths, cov_graph, shortest_graph);
+  return transformSolutionToRoute(v_route, swaths, cov_graph, shortest_graph);
+}
+
+F2CRoute RoutePlannerBase::genShortestRoute(const F2CCells& cells,
+    const F2CSwathsByCells& swaths, const F2CPoint& start, const F2CPoint& end,
+    double d_tol) {
+  F2CRoute route;
+  f2c::rp::RoutePlannerBase route_planner;
+  auto g = route_planner.createShortestGraph(cells, swaths, d_tol);
```
Gonzalo-Mier, line 38: Creating this graph is computationally expensive. Receive one shortest graph as a parameter instead

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -125,19 +142,37 @@ F2CGraph2D RoutePlannerBase::createCoverageGraph(
...
+  // todo put in function
+  F2CPoint deposit_start(-1e8, -1e8);  // Arbitrary point
+  if (this->r_start) {
+    deposit_start = *this->r_start;
+  }
+
+  for (auto&& swaths : swaths_by_cells) {
+    for (auto&& s : swaths) {
+      if (this->r_start) {
+        g.addEdge(s.startPoint(), deposit_start, shortest_graph);
```
Gonzalo-Mier, line 154: Use addDirectedEdge to prevent obtaining a route that starts from the end

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -125,19 +142,37 @@ F2CGraph2D RoutePlannerBase::createCoverageGraph(
...
+    deposit_end = *this->r_end;
   }
 
   for (auto&& swaths : swaths_by_cells) {
     for (auto&& s : swaths) {
-      if (this->r_start_end) {
-        g.addEdge(s.startPoint(), deposit, shortest_graph);
-        g.addEdge(s.endPoint(), deposit, shortest_graph);
+      if (this->r_end) {
+        g.addEdge(s.startPoint(), deposit_end, shortest_graph);
```
Gonzalo-Mier, line 171: Directed edges also

src/fields2cover/utils/parser.cpp
```diff
@@ -103,46 +105,97 @@ F2CCell getCellFromJson(const json& imported_cell) {
...
+  for (auto&& imported_swath : data["features"]) {
     F2CLineString line;
     for (auto&& ps : imported_swath["geometry"]["coordinates"]) {
       line.addPoint(getPointFromJson(ps));
     }
-    swaths.emplace_back(line,
-      imported_swath["properties"]["width"],
-      imported_swath["properties"]["path_id"]);
+    if (!imported_swath["properties"].contains("width")) {
+      std::cout << "Warning: Width property is missing for imported swath, "
```
Gonzalo-Mier, line 167: If you want to put warnings, make sure there is an easy way to deactivate them

src/fields2cover/utils/parser.cpp
```diff
@@ -103,46 +105,97 @@ F2CCell getCellFromJson(const json& imported_cell) {
...
+                   "defaulting to 0.1"
+                << std::endl;
+    }
+
+    auto width = imported_swath["properties"].contains("width")
+                     ? imported_swath["properties"]["width"].get<double>()
+                     : 0.1;  // do not fail if the width is not present
```
Gonzalo-Mier, line 174: This should be a flag. What if the user want to make sure the width is imported and something is failing?

Gonzalo-Mier (changes requested):
I'm missing tests on the new additions

Dockerfile
tomschut: because the vscode devcontainer can set it

.devcontainer/Dockerfile
tomschut: it uses different stages to run.. so the base one runs from build, while the devcontainer one runs from dev. it prevents duplication

src/fields2cover/utils/parser.cpp
tomschut: can you give an example of how to do that in c++? creating a flag?

Dockerfile
Gonzalo-Mier: Is there a problem if vscode set them? (I really don't know, just asking for knowledge)

tomschut: It makes sure that the local files aren't chowned by the dev container. That was an issue.

src/fields2cover/route_planning/route_planner_base.cpp
Gonzalo-Mier: You can create it outside this function, and receive it as an argument of the function.
This function is quite useful, but I'm concerned it will be used iteratively many times, requiring to create the same graph many times, which is expensive. 

As an additional note: If you want to modify the `swaths` to create different graphs, it's faster to save an original graph with all the edges, copy the graph and remove the edges you don't want from that copy.

src/fields2cover/utils/parser.cpp
Gonzalo-Mier: The flag `ALLOW_PARALLELIZATION` is set on the cmake:
https://github.com/Fields2Cover/Fields2Cover/blob/26746b485d09ce41826344a3819e2f87b54cb88e/CMakeLists.txt#L13
And used it on C++ as:
https://github.com/Fields2Cover/Fields2Cover/blob/26746b485d09ce41826344a3819e2f87b54cb88e/src/fields2cover/swath_generator/brute_force.cpp#L8-L10
https://github.com/Fields2Cover/Fields2Cover/blob/26746b485d09ce41826344a3819e2f87b54cb88e/src/fields2cover/swath_generator/brute_force.cpp#L38-L43

src/fields2cover/utils/parser.cpp
Gonzalo-Mier: Here I mean a different thing with flag as in the previous comment. Here I would say to add an argument to the function called `fail_without_swath_width`, or something like this, and it should have a default value of false.

Dockerfile
Gonzalo-Mier: Thank you for the explanation

src/fields2cover/route_planning/route_planner_base.cpp
tomschut: I want to expose it as a top level function, so I'm still not sure how I would create the graph before calling it

Gonzalo-Mier: I guess it's fine then

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -125,52 +159,52 @@ F2CGraph2D RoutePlannerBase::createCoverageGraph(
...
+  g = addStartEndToGraph(cells, swaths_by_cells, shortest_graph, g, deposit_end,
+      this->r_end != std::nullopt);
+
   return g;
 }
 
 std::vector<int64_t> RoutePlannerBase::computeBestRoute(
     const F2CGraph2D& cov_graph, bool show_log) const {
-  int depot_id = static_cast<int>(cov_graph.numNodes()-1);
+  int depot_id = static_cast<int>(cov_graph.numNodes() - 1);
```
Gonzalo-Mier, line 182: As you have added start and end points, now `depot_id` (the starting point) is not always located on the last point. If start and end are set, then `depot_id = cov_graph.numNodes()-2`

src/fields2cover/utils/parser.cpp
```diff
@@ -103,46 +105,97 @@ F2CCell getCellFromJson(const json& imported_cell) {
   return cell;
 }
 
+std::string gen_random(const int len) {
```
Gonzalo-Mier, line 108: Functions should be on camelCase

Gonzalo-Mier: I like this function, but it should go in the Random class as getRandomString

src/fields2cover/route_planning/route_planner_base.cpp
Gonzalo-Mier: Can you add tests on the route planner that show what happens if you set only the start, only the end, and what happens when start and end are extreme points of some swaths?

src/fields2cover/route_planning/route_planner_base.cpp
```diff
@@ -85,15 +104,30 @@ F2CGraph2D RoutePlannerBase::createShortestGraph(
   return g;
 }
 
+F2CGraph2D addStartEndToGraph(const F2CCells& cells,
+    const F2CSwathsByCells& swaths_by_cells, F2CGraph2D& shortest_graph,
+    F2CGraph2D g, F2CPoint deposit, bool point_set) {
```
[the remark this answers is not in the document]
tomschut: I'm not quite sure what you mean by the latter. the point set refers to whether the start and end points are set.. other point is fixed

src/fields2cover/route_planning/route_planner_base.cpp
tomschut: I've added the code, but am unsure about the test you want added here.

Gonzalo-Mier: I've added a new branch called pr-177 with the test I was referring to. It crashes.

src/fields2cover/route_planning/route_planner_base.cpp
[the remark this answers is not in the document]
Gonzalo-Mier: Ok. Understood. I wasn't grasping what point_set meant. :)

src/fields2cover/route_planning/route_planner_base.cpp
Gonzalo-Mier: Once that is solved, we can merge it.

PR #186, opened 2024-11-21
Fix v1.2.1 for ROS Jazzy/Rolling

adivardi:
Currently Nav2 coverage planning still requires `v1.2.1`. Until it is moved to `v2.0.0`, which requires some effort, F2C cannot be used on ROS Jazzy/rolling.
In this PR I have fixed the build issues for this repo. I suggest making a new branch `v1.2.1-devel` and make this PR into that branch (instead of `main` which is at `v.2.0.0`).

This change was tested on Ubunu 24.04 with both Jazzy and rolling. It is ready, but I leave it as draft as the target branch should be changed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(BUILD_PYTHON "Build Python SWIG module" OFF)
 option(BUILD_DOC "Build Documentation" OFF)
 option(BUILD_SHARED_LIBS "Build shared library(.so)" ON)
 
+add_compile_options(-Wno-deprecated-declarations)
 
 
 if (BUILD_PYTHON)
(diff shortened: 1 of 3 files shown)


adivardi:
@Gonzalo-Mier What do you think? Does that sound reasonable?

Gonzalo-Mier:
I guess it makes sense. Please, also note that support from our side on v1.2.1-devel will be limited.

Btw, the conflict on `include/fields2cover/types/MultiLineString.h` leads to a memory leak on v1.2.1.

adivardi:
yeah, I think that is fine. I mostly want to have it running.
Could you please create a new branch from `v1.2.1`? I cannot do that myself.

>Btw, the conflict on include/fields2cover/types/MultiLineString.h leads to a memory leak on v1.2.1.

Oh, good to know. I will fix that then

Gonzalo-Mier:
Thanks for the PR

Gonzalo-Mier:
> let's wait for the CI test to finish as well

Good tip. I'll wait. Thank you

PR #187, opened 2024-11-21
fix build on rolling

adivardi:
Fix an error in the memory leak fix, causing build to fail.

Sorry about that, I didn't clean build this last minute change and somehow missed it.


Gonzalo-Mier:
Thank you again!

PR #190, opened 2024-12-13
Rework swath generator to support filling excess area

pwkamp:
- fields2cover supports overlapping not all or only at the end of the path. Now, this change adds support for 4 different methods of how the excess area should be fille in.
  - NO_OVERLAP: Leave the excess area and do nothing.
  - END_OVERLAP: Add an extra pass at the end that overlaps the previous swath.
  - MIDDLE_OVERLAP: Add an extra pass to the middle and any middle swath will overlap (excludes outside swaths so that the outside swaths stay consistent with the desired width).
  - EVENLY_DISTRIBUTED_OVERLAP: Distribute the overlap evenly to all the swaths, including the outside swaths.
  - Remove allowOverlap parameter as it is redundant with NO_OVERLAP

### DIFF
--- a/include/fields2cover/swath_generator/swath_generator_base.h
+++ b/include/fields2cover/swath_generator/swath_generator_base.h
@@ -14,10 +14,18 @@
 
 namespace f2c::sg {
 
-enum class SwathOverlapType {NO_OVERLAP, END_OVERLAP, MIDDLE_OVERLAP, EVENLY_DISTRIBUTED_OVERLAP};
+enum class SwathOverlapType {
+    NO_OVERLAP = 0,
+    END_OVERLAP = 1,
+    MIDDLE_OVERLAP = 2,
+    EVENLY_DISTRIBUTED_OVERLAP= 3
+};
 
 class SwathGeneratorBase {
  public:
+  SwathOverlapType getOverlapType() const;
+  void setOverlapType(SwathOverlapType);
+
   SwathOverlapType getOverlapType() const;
   void setOverlapType(SwathOverlapType);
 
--- a/include/fields2cover/swath_generator/swath_generator_base.h
+++ b/include/fields2cover/swath_generator/swath_generator_base.h
@@ -23,6 +23,9 @@ enum class SwathOverlapType {
 
 class SwathGeneratorBase {
  public:
+  bool getAllowOverlap() const;
+  void setAllowOverlap(bool);
+
   SwathOverlapType getOverlapType() const;
   void setOverlapType(SwathOverlapType);
 
--- a/include/fields2cover/swath_generator/swath_generator_base.h
+++ b/include/fields2cover/swath_generator/swath_generator_base.h
@@ -14,10 +14,12 @@
 
 namespace f2c::sg {
 
+enum class SwathOverlapType {NO_OVERLAP, END_OVERLAP, MIDDLE_OVERLAP, EVENLY_DISTRIBUTED_OVERLAP};
+
 class SwathGeneratorBase {
  public:
-  bool getAllowOverlap() const;
-  void setAllowOverlap(bool);
+  SwathOverlapType getOverlapType() const;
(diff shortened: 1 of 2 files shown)


pwkamp:
Bare with me as this is the first time I've done a PR....

I took inspiration from comments [PR-168](https://github.com/Fields2Cover/Fields2Cover/pull/168) and I think this should implement the comments and ideas made in that thread.

Gonzalo-Mier:
Hi @pwkamp, thanks for the PR. I love your variable names. Let me review the PR :)

include/fields2cover/swath_generator/swath_generator_base.h
```diff
@@ -14,10 +14,12 @@
 
 namespace f2c::sg {
 
+enum class SwathOverlapType {NO_OVERLAP, END_OVERLAP, MIDDLE_OVERLAP, EVENLY_DISTRIBUTED_OVERLAP};
```
Gonzalo-Mier, line 17: Add int to each type, so it'll prevent future errors if someone use the indexs to refer to the SwathOverlapType and we remove any of the methods. So something like:

```
enum class SwathOverlapType {
  NO_OVERLAP = 0,
  END_OVERLAP = 1,
  MIDDLE_OVERLAP = 2,
  EVENLY_DISTRIBUTED_OVERLAP = 3,
};
```

include/fields2cover/swath_generator/swath_generator_base.h
```diff
@@ -14,10 +14,12 @@
...
+
 class SwathGeneratorBase {
  public:
-  bool getAllowOverlap() const;
-  void setAllowOverlap(bool);
```
Gonzalo-Mier, line 20: Do not remove functions that people may be using. If they pull the changes, they will also have to change their code, and we don't want to give more work to users than needed.

For the get, I would only return false if it's `NO_OVERLAP`. For the set, if the argument is false, then `NO_OVERLAP`, and if true, I would pick `EVENLY_DISTRIBUTED_OVERLAP`

src/fields2cover/swath_generator/swath_generator_base.cpp
```diff
@@ -48,12 +49,45 @@ F2CSwaths SwathGeneratorBase::generateSwaths(double angle,
   F2CPoint min_point(rot_poly.getDimMinX(), rot_poly.getDimMinY());
   auto seed_curve = rot_poly.createStraightLongLine(min_point, 0.0);
 
-  double curve_y {-0.5 * op_width};
   F2CMultiLineString paths;
-  while (field_height > curve_y + (allow_overlap ? 0.0 : 0.5 * op_width)) {
-    curve_y += op_width;
+
+  double increment = op_width;
```
Gonzalo-Mier, line 54: Is there a reason for using increment instead of op_width? And for putting this line here instead of line 81?

src/fields2cover/swath_generator/swath_generator_base.cpp
```diff
@@ -48,12 +49,45 @@ F2CSwaths SwathGeneratorBase::generateSwaths(double angle,
...
+  int num_paths = std::floor(field_height / op_width);
```
Gonzalo-Mier, line 55: Why num_paths is not on line 81?

src/fields2cover/swath_generator/swath_generator_base.cpp
```diff
@@ -48,12 +49,45 @@ F2CSwaths SwathGeneratorBase::generateSwaths(double angle,
...
+  double beginning_offset = 0.5 * op_width;
```
Gonzalo-Mier, line 56: move this variable near where it is used

Gonzalo-Mier (changes requested):
I like the implementation and think people can greatly benefit with this. Sadly, I'm not that sure if this works as intended. Can you add some tests that try the new methods and check if the swaths are placed correctly?

Issue #195, opened 2025-03-20
CMake standard for BUILD_TESTING


impala454:
Hi I was going to just submit a PR for this but curious if there was specific reasoning first.

The cmake for this project adds its own special

```option(BUILD_TESTS "Build tests" ON)```

instead of using the [standard cmake built in option](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html) `BUILD_TESTING`.  This results in Fields2Cover not obeying standard build commands like colcon mixins or any other scripts which turn off `BUILD_TESTING`.  Would it be fine to remove this "custom" option and just roll with the cmake standard option which most other ROS packages use?

Gonzalo-Mier:
Hi @impala454, 

No specific reason. I probably copied it from somewhere else. Feel free to make the PR. Nice that you realised!

PR #196, opened 2025-03-21
Update cmake testing option to standard

impala454:
Minor update to make the testing cmake option conform to the cmake standard so tools like colcon will properly enable/disable testing with mixins.  Addresses #195.


Gonzalo-Mier:
Thank you for the PR

Issue #201, opened 2025-04-14
Boustrpheodon cell decomposition does not work with linearRings that have horizontal edges


lyfkyle:
I found that the BCD implementation in F2C creates too many cells when the cell linearRings have horizontal edges (after rotation). 

This is my test image.
![Image](https://github.com/user-attachments/assets/e930f3ca-2ccc-4751-89ed-a2cb3a3edc67)

And the output from BCD is:
![Image](https://github.com/user-attachments/assets/f152f5ed-3a65-4e8c-93f9-d4f082088b9b)

The culprit is this line https://github.com/Fields2Cover/Fields2Cover/blob/main/src/fields2cover/decomposition/boustrophedon_decomp.cpp#L19
```       
if (line1.length() > 1e-5 && line2.length() > 1e-5) {
    lines.addGeometry(line1);
    lines.addGeometry(line2);
}
```

Because there are horizontal edges, line1 and line2 both have non-zero lengths despite not actually going out of the linearRing. 

To fix, we can check whether line is an edge on the linearRing and skip if necessary.
```
auto line1 = cell.createLineUntilBorder(p, split_angle);
while (isLinePartOfLinearRing(ring, line1)) {
    line1 = cell.createLineUntilBorder(line1.endPoint(), split_angle);
}
auto line2 = cell.createLineUntilBorder(p, split_angle + M_PI);
while (isLinePartOfLinearRing(ring, line2)) {
    line2 = cell.createLineUntilBorder(line2.endPoint(), split_angle + M_PI);
}
```

The output with this change
![Image](https://github.com/user-attachments/assets/2d79a897-64e2-4317-8b0b-15827a440a25)

Raising a PR. Let me know your thoughts.

Gonzalo-Mier:
Welcome to the project and thank you for the issue and the PR. Nice job!

PR #202, opened 2025-04-14
Fix BCD with horizontal edges

lyfkyle:
Fix BCD with horizontal edges #201


Gonzalo-Mier:
Nice! Thank you really much @lyfkyle

lyfkyle:
Is there a plan to release to ROS2 humble?

PR #204, opened 2025-04-22
Feature/expose or tools, add curvature constrained curves

kylona:
Exposes some of the F2C internals to the python bindings so Quickplan can make use of curvature constrained turns, and can use the computeBestRoute directly.
int64_t needed to be redefined as long long int for swig to be happy with it. 
Use the GUIDED_LOCAL_SEARCH metaheuristic instead of "AUTOMATIC" because the automatic search heuristic is just a greedy search which did not work well for our use cases. 
For the computeBestRoute expose a time_limit_seconds input to give python better control of the search. 

Not sure if these changes are something that should be propagated to the root repository or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,22 +1,13 @@
 
-set(FIELDS2COVER_MAJOR_VERSION "2")
-set(FIELDS2COVER_MINOR_VERSION "0")
-set(FIELDS2COVER_PATCH_VERSION "0")
+cmake_minimum_required(VERSION 3.12.4...3.22)
 
-set(FIELDS2COVER_VERSION_STRING
-  "${FIELDS2COVER_MAJOR_VERSION}.${FIELDS2COVER_MINOR_VERSION}.${FIELDS2COVER_PATCH_VERSION}"
-)
-set(FIELDS2COVER_VERSION
-  ${FIELDS2COVER_MAJOR_VERSION}.${FIELDS2COVER_MINOR_VERSION}.${FIELDS2COVER_PATCH_VERSION}
+project(Fields2Cover
+  VERSION 2.0.0
+  DESCRIPTION "Fields2Cover is a complete coverage path planning package for autonomous robots"
+  HOMEPAGE_URL "http://fields2cover.github.io"
+  LANGUAGES CXX
 )
 
-set(PACKAGE_NAME "Fields2Cover")
-set(PACKAGE_DESCRIPTION
-  "Fields2Cover is a complete coverage path planning package for autonomous robots"
-)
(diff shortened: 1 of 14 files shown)


Gonzalo-Mier:
Hi @kylona, thank you for the PR. I couldn't fix it, so thank you for solving it. Just a few things to merge them:

- Why do you use `long int` instead of `long long int` as you mentioned in the text?
- In src/fields2cover/path_planning/dubins_curves_cc.cpp, keep the same coding style as the repository
- I prefer that choosing GUIDED_LOCAL_SEARCH is a boolean choice in the function.

kylona:
Hi @Gonzalo-Mier 

I'm glad you are open to these changes. 
But I need to say "oops" because I intended to pull this into main of the fireflyautomatix fork of f2c. 

> Not sure if these changes are something that should be propagated to the root repository or not.

Although what I meant by this was if we should also submit here. I guess the answer is yes.

When this was just for our fork I was happy to make opinionated changes tailored to our use case, but you are right best to have flags for the general public to decide what they want. 

I also think this deserves 2 PRs. One for exposing curvature-constrained curves and one for exposing or-tools. Unless you would rather keep them in one. What do you prefer? 

I need to look at `long int` vs `long long int` to make sure `long long int` works ok. I think I couldn't keep track in my head which was how many bits. If `long long int` builds ok that's preferred. Although I suspect that no user of f2c is likely to use more than 2^31 nodes so if I have too much trouble I may leave it as is. 

I'll have an edit soon.

Gonzalo-Mier:
> I also think this deserves 2 PRs. One for exposing curvature-constrained curves and one for exposing or-tools. Unless you would rather keep them in one. What do you prefer?

Whatever you prefer. I'm happy with both.

> I need to look at long int vs long long int to make sure long long int works ok. I think I couldn't keep track in my head which was how many bits. If long long int builds ok that's preferred. Although I suspect that no user of f2c is likely to use more than 2^31 nodes so if I have too much trouble I may leave it as is.

I'm not afraid with the number of nodes. The problem is the cost of the connections. In or-tools they have an Infinite value which is 2^63-1 to mark that an edge is not connected.

Gonzalo-Mier:
Also note that I'm working on releasing the v3.0, which is in the pre-v3.0 branch. Still a lot of work to do, but just in case you want to start checking it out.

kylona:
Ok. Got those changes made. I decided to just leave them together and I updated the title to indicate the whole contents. 

I'll check V3 out! V2 added a ton of nice functionality so I'm excited to find out what's up your sleeve.

Gonzalo-Mier:
Thank you a lot for the PR!

PR #219, opened 2026-05-20
Improvement: reduce memory usage

icpmoles:
Hi
Our group was trying to use the library with really big fields (think around 1k swaths with a really detailed ring geometry) and struggled with the memory usage.

We are proposing the following change:
- instead of pre-calculating the shortest distances and the shortest paths between all nodes, only calculate the shortest distances to feed to or-tools and then reconstruct the shortest paths only when needed to recreate the route.
  - a further improvement would be skipping the route reconstruction in the `transformSolutionToRoute` altogether: it's only used when interconnecting the swaths at the route level, but when converting into a path the solution relies only on splines. Probably breaking so it's not included in this PR.
  - another option is to have an option to directly convert a route into a path by connecting the swaths with the connections obtained from the graph, probably useful only for small UGVs that don't need smooth trajectories.


icpmoles:
Did they change anything upstream on how colcon resolves dependencies?

I'm able to complete the ROS2 check with both jazzy and kilted, but it fails with rolling.

EDIT:
Now I'm even more confused:

with `rostooling/setup-ros-docker:ubuntu-noble-ros-rolling-ros-base-latest` I'm able to build jazzy and kilted.

switching to `rostooling/setup-ros-docker:ubuntu-resolute-ros-rolling-ros-base-latest` I'm able to build rolling correctly.

I'm unfortunately not really familiar with the ROS CI infrastructure.

Gonzalo-Mier:
Hi @icpmoles, welcome to the Fields2Cover project as a collaborator! I have seen that you made a tutorial of Fields2Cover and we are greatful for it <3.

Don't worry about ROS2. The ortools_vendor we also provide is failing lately due to a deprecation in a third party library. During this period I'm a little busy to fix it, but it is in my todo list. 

About the PR, I like the idea of it and I think it could be beneficial for all the users of the project. We are right now working on merging the branch v3 to main, so it would be awesome that the changes were made over that branch instead of main directly. Moreover, I would also like to see some numbers of how much the PR improves. Could you measure the memory and the time consumed before and after the PR?
One last thing. I have seen that in a few functions the output parameter was changed to void. Despite it may be a good idea, we would like to keep the ABI of the functions as stable as possible during major versions.

Again, thank you a lot for your contribution!

icpmoles:
I ran some tests with Valgrind Massif on this map 
[194.json](https://github.com/user-attachments/files/28072725/194.json), 231 swaths

Upstream 2.0
upstream2 0

My PR
proposal

In the first picture, if you notice the spike it's due to the compiler not optimizing the copy elision for the return inside `Graph::shortestPathsAndCosts(int64_t INF)` for some reason and causing an useless copy that gets discarded immediately (I tried with both gcc and clang but neither of them was able to optimize the call).

I'm aware it's breaking, I couldn't figure out a more elegant solution unfortunately.

For the time difference I can send the results tomorrow.

I'll check the v3 branch too.

Gonzalo-Mier:
It looks awesome! Can you increment the version by one? So if main is 2.0.0 -> 2.1.0. Same with v3 if you change that one.

icpmoles:
Sorry for the late reply. These are the new timings in seconds:


original v2
createShortestGraph: 8.08101
createCoverageGraph: 7.04742
transformSolutionToRoute: 0.835381



pr
createShortestGraph: 7.98346
createCoverageGraph: 6.41849
transformSolutionToRoute: 0.83789


Take them with a grain of salt because of thermal throttling.

Gonzalo-Mier:
Thank you really much for your contribution. You made Fields2Cover a little bit better :)

PR #222, opened 2026-07-10
Fix segfault in FieldCoverage::computeCost when swaths is empty

### LINKED ISSUE #221: Segfault due to small field zero-swath case
https://github.com/Fields2Cover/Fields2Cover/blob/0ce7fcce82f3f79f773699448769bcd679c739bc/src/fields2cover/objectives/sg_obj/field_coverage.cpp#L16-L26

I've got a tiny little bug I found while using fields2cover in a small industrial environment. Basically, if the field to cover is too small and headland takes up basically all available space, we might end up in a scenario where no swath can be generated. See the link above to view the culprit code.

What I did to fix it in my fork was to simply set the cost to 0.0 if the size of the swaths variable was less than 1. My rationale was that this is still a valid guess, so we don't want to throw any exceptions, but clearly this swath angle was very poor so cost should be minimal.

I can make a PR with the small patch fix, but I first just wanted to clarify that this would be the right solution, and not something different upstream of the call.

SAY-5:
Closes #221.

When a field is small enough that the headland consumes nearly all the area, the swath generator can return zero swaths. `FieldCoverage::computeCost` then reads `swaths[0].getWidth()` on line 24, which indexes past the end of the swath vector and segfaults.

This returns a cost of 0.0 when there are no swaths, matching the issue author's reasoning: no swaths cover no area, so the coverage ratio is 0. Added a regression test that exercises the empty-swaths path against both `computeCost` overloads.

### DIFF
--- a/src/fields2cover/objectives/sg_obj/field_coverage.cpp
+++ b/src/fields2cover/objectives/sg_obj/field_coverage.cpp
@@ -15,6 +15,10 @@ double FieldCoverage::computeCost(
 
 double FieldCoverage::computeCost(
     const F2CCells& poly, const F2CSwaths& swaths) {
+  if (swaths.size() == 0) {
+    return 0.0;
+  }
+
   F2CMultiLineString lines;
   for (const auto& s : swaths) {
     lines.addGeometry(s.getPath());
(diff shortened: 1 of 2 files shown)


SAY-5:
Done, switched those six assertions to EXPECT_NEAR with a 1e-5 tolerance so the epsilon in the denominator doesn't trip exact equality.

Gonzalo-Mier:
Perfect, thank you really much for your PR!

PR #205, opened 2025-05-02
Demonstrate Mac OSX build

rodja:
This PR demonstrates a successful build on Mac OSX with Silicon (arm64) as a step towards https://github.com/Fields2Cover/Fields2Cover/issues/26. This PR is not intended to be merged. It just shows the hacky code changes I need to perform in addition to the following install commands:

```bash
brew install tinyxml2 gdal swig geos
virtualenv .venv
source .venv/bin/activate
pip install mypy-protobuf mypy
cmake -DBUILD_PYTHON=ON -DBUILD_TUTORIALS=OFF -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_DOC=OFF -DUSE_ORTOOLS_FETCH_SRC=ON -DPython3_EXECUTABLE=$(which python) -DCMAKE_PREFIX_PATH="$(brew --prefix gdal)" ..
make -j$(nproc)
sudo make install
```

PR #225, opened 2026-07-13
Include GDAL headers without the distro-specific gdal/ prefix

pascalzauberzeug:
Headers include `<gdal/ogr_geometry.h>` etc. The `gdal/` subdirectory is a Debian/Ubuntu packaging detail — GDAL installs its headers flat, so on macOS (Homebrew) and other layouts the prefixed form isn't found. Since these headers are installed, this also breaks downstream C++ consumers.

Fix: use the prefix-free form. The include directory is already provided by the linked `GDAL::GDAL` target on every platform (on Ubuntu it points at `/usr/include/gdal`, so the prefix-free form resolves there too). This also lets the "Configure GDAL" CI step — which manually recreated the `gdal/` layout inside the `osgeo/gdal` containers — be removed, so the same CI now proves the fix.

Related to #205.

<details><summary>Note for non-CMake consumers</summary>

Non-CMake consumers on Debian who relied on the `gdal/` prefix now need `-I/usr/include/gdal` on the compiler line. CMake consumers get it automatically from `GDAL::GDAL`.
</details>

---

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,11 +9,6 @@ LABEL NAME="fields2cover" \
 ENV DEBIAN_FRONTEND noninteractive
 
 WORKDIR /workspaces/
-RUN mkdir -p /usr/include/new_gdal && \
-    cp -r /usr/include/gdal* /usr/include/new_gdal/ && \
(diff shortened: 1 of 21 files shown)


Gonzalo-Mier:
Thank you for another PR!

PR #226, opened 2026-07-13
Find GEOS and TBB via find_package with imported targets

pascalzauberzeug:
The build hard-codes `-lgeos_c` and the bare `tbb` library name, which are only found when the libraries sit on the default linker path — not the case for Homebrew on macOS. Prefer the CMake configs both projects ship (`GEOS::geos_c` since GEOS 3.8, `TBB::tbb` in oneTBB) via `find_package(... CONFIG QUIET)`, falling back to the old bare names where they aren't available.

<details><summary>Config export</summary>

`GEOS::geos_c` is a PUBLIC link dependency, so it ends up in the exported targets. `Fields2CoverConfig.cmake.in` now calls `find_dependency(GEOS CONFIG)` — but only in the branch where the config was actually used at build time, so the plain `-lgeos_c` fallback needs no GEOS package downstream.
</details>

Related to #205.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,7 @@ if(ALLOW_PARALLELIZATION)
   # Prefer the imported target from oneTBB's CMake config; fall back to the
   # bare library name for distros that don't ship the config.
   find_package(TBB CONFIG QUIET)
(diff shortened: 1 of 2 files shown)


Gonzalo-Mier:
Thank you again for another PR!

PR #230, opened 2026-07-20
Update Dockerfile image base and multistage

lcsvillela:
I'm using Fields2Cover and I think it's a good ideia to update the image, I got the last image from GDAL to build and I've used the multistage to have a lower image, because the size was too big


Gonzalo-Mier:
It seems awesome!

Welcome to the Fields2Cover project, @lcsvillela, and thank you for the PR

Issue #231, opened 2026-08-05
simplifyConnection discards the connection polyline based on endpoint distance alone


u33549:
## Summary

`PathPlanning::simplifyConnection` decides whether to follow a connection's waypoints or cut straight across them using the **straight-line distance between the connection's endpoints**, without ever measuring how far the waypoints in between actually depart from that line. When a connection detours around real geometry but happens to start and end close together, the waypoints are discarded and the path is planned straight through the obstacle the connection existed to avoid.

Reported against **v2.0.0**, the released version we depend on. I checked `main` and the `v3.0` branch as well — the condition is byte-identical in all three, so this is not something an upgrade resolves.

## Where

`src/fields2cover/path_planning/path_planning.cpp`:

```cpp
std::vector<std::pair<F2CPoint, double>> PathPlanning::simplifyConnection(
    const F2CRobot& robot,
    const F2CPoint& p1, double ang1,
    const F2CMultiPoint& mp,
    const F2CPoint& p2, double ang2) {
  const double safe_dist = getSmoothTurningRadius(robot);
  std::vector<std::pair<F2CPoint, double>> path;
  path.emplace_back(p1, ang1);

  if (p1.distance(p2) < 6.0 * safe_dist || mp.size() < 2) {
    path.emplace_back(p2, ang2);   // <-- mp is discarded entirely
    return path;
  }
  ...
```

`p1.distance(p2)` is the distance from the end of one swath to the start of the next. `mp` is the connection the route planner produced to get between them. The test relates the two endpoints to each other, but nothing in it looks at `mp`.

## Why this is a problem

The two quantities are independent. A connection routed around a concave notch, or between two cells of a decomposed field, regularly has endpoints a few meters apart while the waypoints between them run for tens of meters around the boundary. Whenever that separation falls under `6 * safe_dist`, `mp` is dropped and `createTurn` is asked for a direct maneuver between the two poses — straight across the ground the route deliberately went around.

A test on the deviation of `mp` from the segment `p1`–`p2` would separate the two cases. The endpoint distance cannot, because a long detour and a short hop are indistinguishable by that measure.

## Second issue in the same function

The corner test does not account for angle wrapping:

```cpp
double ang_in  = (mp[i] - mp[i-1]).getAngleFromPoint();
double ang_out = (mp[i+1] - mp[i]).getAngleFromPoint();
if (fabs(ang_in - ang_out) > 0.1) {
  vp.emplace_back(mp[i]);
}
```

[...]

PR #233, opened 2026-08-14
Small API additions: Route::reverse, TurningBase::hasContinuousCurvature, public getSmoothTurningRadius

u33549:
### Motivation

Three pieces of information the library already has internally but does not offer to
callers: the radius a continuous-curvature turn sweeps, whether a turn planner keeps the
curvature continuous, and reversing a route.

Each one currently forces the caller into something worse — duplicating a formula it
cannot compute exactly, branching on concrete planner types, or rebuilding a `Route` by
hand. All three are additive, none changes behavior, and each is one commit.

They are prerequisites for the `planPathForConnection` fix (#231), but they are gaps in
their own right, and sending them first keeps that fix reviewable as a single concern.

### 1. Make `PathPlanning::getSmoothTurningRadius` public

Moved the declaration from the private block to the public one. Body untouched.

The value cannot be computed outside the library: the function calls `end_of_clothoid`
from `steering_functions/utilities/utilities.hpp`, and those headers are not part of the
install (only `fields2cover` and `matplot` land in the install prefix). We currently
approximate it downstream with a different formula, which drifts from what F2C itself
plans.

### 2. Add `TurningBase::hasContinuousCurvature()`

Added `virtual bool hasContinuousCurvature() const` returning `false`, overridden to
`true` in `DubinsCurvesCC` and `ReedsSheppCurvesHC`.

Choosing the radius to round a corner with depends on whether the turn planner ramps the
curvature or steps it. A caller holding a `TurningBase&` has no way to ask, so it would
have to branch on the concrete type and silently pick wrong for any planner added later.

Querying the robot instead is not equivalent: `tutorials/6_path_planning.cpp` drives
plain `DubinsCurves` with a robot that has `MaxDiffCurv` set. Continuity is a property of
the planner, not of the robot.

Left non-pure so that `TurningBase` subclasses outside this repo keep compiling.

### 3. Add `Route::reverse()`

Reverses the group order, each `Swaths`, each `Swath`, and the point order of each
connection.

`Swath::reverse()` and `Swaths::reverse()` already exist — `Route` was the missing one,
so anything that needs a route driven backwards has to rebuild it by hand.

One detail worth a look during review: connections are stored *before* the swaths they
lead into and the trailing one is optional, so `connections_.size()` is either
`v_swaths_.size()` or one more. Reversing both vectors on their own shifts the
connections by one whenever there is no trailing connection. The implementation appends
an empty connection first when needed and drops a trailing empty one afterwards, so
reversing twice returns the exact original, connection count included.

Two smaller things: `Swaths::reverse()` only reverses the swath order and not the driving
direction of each swath, so `Swath::reverse()` is called per swath as well; and
`MultiPoint` has no `reversePoints()` (unlike `LineString`/`LinearRing`), so connections
are rebuilt.

### Note

Adding a virtual method changes the vtable layout, so this is an ABI break.

### DIFF
--- a/include/fields2cover/path_planning/dubins_curves_cc.h
+++ b/include/fields2cover/path_planning/dubins_curves_cc.h
@@ -19,6 +19,7 @@ class DubinsCurvesCC : public TurningBase {
   F2CPath createSimpleTurn(const F2CRobot& robot,
       double dist_start_pos, double start_angle,
       double end_angle) override;
+  bool hasContinuousCurvature() const override;
   F2CPath createConstrainedTurn(const F2CRobot& robot,
       const F2CPoint& start_pos, double start_angle, double start_curvature,
       const F2CPoint& end_pos, double end_angle, double end_curvature, const bool fix_loops=false);
--- a/include/fields2cover/path_planning/path_planning.h
+++ b/include/fields2cover/path_planning/path_planning.h
@@ -40,8 +40,11 @@ class PathPlanning {
       const F2CPoint& p2, double ang2,
       TurningBase& turn);
 
- private:
+  /// Radius swept by a turn with continuous curvature. Wider than the minimum
+  /// turning radius, as the clothoid ramps spend part of the deflection.
   static double getSmoothTurningRadius(const F2CRobot& robot);
+
+ private:
   static std::vector<std::pair<F2CPoint, double>> simplifyConnection(
       const F2CRobot& robot,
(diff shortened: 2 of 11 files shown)


u33549:
@Gonzalo-Mier Is there anything need to change here

Gonzalo-Mier:
Hi @u33549, Thank you for the PR! Seems great

PR #234, opened 2026-08-18
Follow connection polylines instead of cutting across them

u33549:
### Motivation

`planPathForConnection()` ignores the polyline it is given. It hands every consecutive pose
pair to the turn planner, so a connection detouring around a concave notch or between
decomposed cells is driven straight from one end to the other, cutting across ground the
route deliberately routed around. `simplifyConnection()` also compares headings with
`fabs(ang_in - ang_out)` on raw angles, which wraps wrong.

The library's own test says as much: `planPathForConnection` has been `DISABLED_` since
v2.0.0's first commit, and forcing it to run shows the path coming out shorter than the
track it is meant to follow.

`PathSectionType::HL_SWATH` is equally unfinished — never produced or consumed anywhere,
skipped by `discretizeSwath()`, unreachable through any public API. It is in this PR because
the fix produces it: the straight legs between rounded corners are headland passes, not
turns, and without the `HL_SWATH` half each would be a type `discretizeSwath()` refuses to
subdivide, leaving it a single waypoint.

### 1. Discretize `HL_SWATH` sections (`83dc019`)

`discretizeSwath()` only densified `SWATH` states. One condition, plus the doc comment.

No call path in the tree changes behaviour, because nothing produced the type.

### 2. Let `appendSwath` write the section type (`e9a18f1`)

```cpp
void appendSwath(const Swath& swath, double cruise_speed);                        // existing
void appendSwath(const Swath& swath, double cruise_speed, PathSectionType type);  // added
```

The two-argument form delegates with `SWATH`. An overload rather than a default argument,
so the ABI is untouched.

`setSwathType()`/`setTurnType()` already exist and have no callers in the library — they
are there for the outside, and the third type had no counterpart.

If the preferred answer is to drop the enumerator instead: it is serialized as `3` by
`serializePath`, so removing it changes the file format, and telling a headland pass from
a swath is a real need — `generateHeadlandSwaths` returns rings, and turning one into a
drivable path is the next thing we need from the library.

### 3. Follow and round the connection track (`9c3a06e`)

`planPathForConnection` now walks the polyline: straight runs become `HL_SWATH` legs,
and only genuine corners go to `createTurn`. Corners too close together to take one at a
time are folded into a single maneuver; a corner a turn could only round by cutting past
`cut_tol` is left square instead. A connection whose track stays within a quarter of the operation width
of the straight hop, or one that reverses over a short hop, is still driven direct — the
u-turn between neighbouring swaths is the hop itself, and an arc through a 180-degree
vertex has no tangent length to round with.

Four details decide whether that produces corners or loops. Each was checked by reverting it
and confirming a test goes red:

- **The approach is sized from the radius the turn actually sweeps** — the wider of
  `getSmoothTurningRadius` (public since #233; the clothoid ramps spend part of the
  deflection) and `sqrt(rate * sweep)` (a shallow corner never reaches full curvature). They
  cross around 50 degrees. Take only the first and a shallow corner starts its turn a metre
  early, cutting ground the covering legs either side of it need.
- **A turn is judged by how far it turns, not how long it is.** A loop's extra `2*pi*R` is a
  couple of metres at a small radius — inside any sane length budget, and smooth enough that
  no heading-step check sees it. Turning does not hide it, and it catches doubling-back S
  shapes too.
- **`cut_tol` is floored in turning radii, capped at the operation width.** Tied to the
  width alone, an implement whose turning circle is close to its own width gets every corner
  refused; the cap stops a wide-turning one cutting corners it should leave square.
- **The tangent margin applies only to a continuous-curvature turn**, being room for the
  clothoid lead-in. A discontinuous turn has nothing to ramp.

`simplifyConnection` is removed, body and private declaration. **This closes the second half
of #231 by construction** — the angle-wrap bug lived there, and the new code uses
`Point::getAngleDiffAbs` throughout.

One more fix along the way: at a pinned route end taking a single corner the outgoing
reference point *is* the corner, so the lateral-offset bound collapsed to zero and refused a
corner the tangent bound allowed.

### 4. Tests (`7f7b322`)

`fields2cover_pp_pp.planPathForConnection` is enabled with its expectations untouched.

Nine new tests over behaviour that had no coverage: detour corners, tight jogs folded into
one S, neighbour u-turns, curved tracks that must not be flattened, corners left sharp
because rounding would cut too much, corners taken from the swath's heading not the track's,
open-ended route connections, shallow-corner approach floors, and corners too close to
round one at a time.

`distanceToPath`, `maxHeadingStep`, `turnStarts` and `maxTurnSweep` go into the checker.
`distanceToPath` measures to the polyline through the states rather than to the states
themselves — a straight run is a single state, so measuring to states alone would miss the
track between them. `maxTurnSweep` is the loop fence: it takes the widest excursion of a
turn's running signed heading change, so a corner or an S reads under half a revolution
while a loop runs past a full one.

The shallow-corner test bounds the approach on *both* sides: a turn that starts at the
corner has no tangent to turn in, and one that starts too far back leaves the leg early and
cuts ground the covering legs either side of it need.

One candidate detail was dropped after the same check — adding the clothoid ramp length to
the approach distance. With the turning fence in place it changed no output at all, and at a
swath junction it made the turn slightly wider.

### Known limitations

The tuning constants (`cut_tol`'s fractions, the u-turn thresholds, the span limit) are
`constexpr` in an anonymous namespace rather than parameters. `f2c::pp::PathPlanning` is
entirely static, so exposing them raises a static-vs-instance question that is wider than
this change. Making `cut_tol` a defaulted argument would be a sensible follow-up: how much
ground a machine may cut across depends on the field and on what the implement is doing, so
it is the one an operator would plausibly want to set.

Where a connection meets a swath, the entry pose is the swath's own end pose and cannot
slide back along it, so a continuous-curvature turn there has no approach room and swings
slightly wide before settling. How wide depends on the curvature rate: it is small at rates
a field machine uses and grows as the rate falls. Bounded and drivable, but not as tight as
a corner with room on both sides.

### DIFF
--- a/src/fields2cover/path_planning/path_planning.cpp
+++ b/src/fields2cover/path_planning/path_planning.cpp
@@ -4,10 +4,308 @@
 //                        BSD-3 License
 //=============================================================================
 
+#include <algorithm>
+#include <array>
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <vector>
 #include <steering_functions/utilities/utilities.hpp>
 #include "fields2cover/path_planning/path_planning.h"
 
 namespace f2c::pp {
+
+namespace {
+
+// Fractions of the operation width, unless noted otherwise.
+constexpr double kLegShare = 0.5;         // a leg is split evenly between its two corners
+constexpr double kMinSweep = 0.05;        // corners under ~3deg are driven straight through
+constexpr double kRadiusMargin = 1.2;     // room for the clothoid lead-in of a CC turn
+constexpr double kMinBackoffRadii = 0.5;  // shallow corners still get a workable approach
+constexpr double kTurnSlack = 1.0;        // turning past this much extra is a loop or an S
+constexpr double kReversalSweep = 2.0;    // legs this far apart in heading double back
+constexpr double kUturnHopWidths = 3.0;   // a u-turn between neighbours spans about one
+constexpr double kUturnReachRadii = 4.0;  // how far back a u-turn may start, in radii
+constexpr size_t kMaxSpan = 3;            // a spur, the corner it leads to, and a spur out
+
+// Below this deviation from the straight hop, a connection is driven direct.
+constexpr double kDirectHopDev = 0.25;
+// Points this close to the chord they'd round are dropped before rounding.
+constexpr double kSimplifyTol = 0.1;
+
+// cut_tol is the wider of these two -- see cutTolerance() for why.
+constexpr double kCutTolWidths = 0.5;
+constexpr double kCutTolRadii = 1.0;
+
+// A leg runs along the connection's own track, driven at cruise speed.
+void addStraight(F2CPath& path, const F2CRobot& robot,
+    const F2CPoint& a, const F2CPoint& b) {
+  const double len = a.distance(b);
+  if (len < 1e-6) {
+    return;
+  }
+  F2CPathState s;
+  s.point = a;
+  s.angle = (b - a).getAngleFromPoint();
+  s.len = len;
+  s.dir = f2c::types::PathDirection::FORWARD;
+  s.type = f2c::types::PathSectionType::HL_SWATH;
+  s.velocity = robot.getCruiseVel();
+  path.addState(s);
+}
+
+// Radius the turn actually sweeps, which is what the approach to a corner has
+// to be sized from. Two effects widen it past the minimum turning radius, and
+// they dominate at opposite ends of the range:
+//
+//   - the clothoid ramps spend part of the deflection, so a turn that reaches
+//     full curvature sweeps wider than 1/kappa (getSmoothTurningRadius);
+//   - a shallow corner never reaches full curvature at all, peaking at
+//     sqrt(rate * sweep), so the gentler the corner the wider it sweeps.
+//
+// Taking the larger keeps the approach honest at both ends. Dropping the
+// deflection term lets a shallow corner be planned from a radius it never
+// drives, which starts the turn too late and overlaps the leg it came in on.
+double cornerRadius(const F2CRobot& robot, double sweep, bool continuous) {
+  const double min_radius = robot.getMinTurningRadius();
+  const double rate = robot.getMaxDiffCurv();
+  if (!continuous || rate <= 0.0) {
+    return min_radius;
+  }
+  double radius = std::max(min_radius, PathPlanning::getSmoothTurningRadius(robot));
+  if (sweep > 0.0) {
+    radius = std::max(radius, 1.0 / std::sqrt(rate * sweep));
+  }
+  return radius;
+}
+
+// How far rounding a corner may leave the track. Floored in turning radii,
+// capped at the operation width: a compact implement (radius close to its own
+// width) still gets room for a fillet, but a wide-turning one doesn't get a
+// tolerance loose enough to cut corners it should instead leave sharp.
+double cutTolerance(const F2CRobot& robot) {
+  return std::max(
+      std::min(kCutTolRadii * robot.getMinTurningRadius(), robot.getCovWidth()),
+      kCutTolWidths * robot.getCovWidth());
+}
+
+// Point `dist` along the ray from `from` to `to`; `from` itself if they coincide.
+F2CPoint pointAlong(const F2CPoint& from, const F2CPoint& to, double dist) {
+  if (from.distance(to) < 1e-9) {
+    return from;
+  }
+  return from.getPointFromAngle((to - from).getAngleFromPoint(), dist);
+}
+
+// Furthest any state of `arc` strays from the polyline it stands in for.
+double deviationFromTrack(const F2CPath& arc, const std::vector<F2CPoint>& track) {
+  double worst = 0.0;
+  for (auto&& s : arc.getStates()) {
+    double nearest = std::numeric_limits<double>::max();
+    for (size_t k = 0; k + 1 < track.size(); ++k) {
+      nearest = std::min(nearest, s.point.distance(F2CLineString(track[k], track[k + 1])));
+    }
+    worst = std::max(worst, nearest);
+  }
+  return worst;
+}
+
+// Follow `poly`: straight runs as they are, corners through the turn planner.
+// `start_angle`/`end_angle` are the swath headings at either end, unset if
+// there's none -- the track's own first/last leg is often a spur off the
+// border graph the robot never drives (see PR-4, B.3). Returns the count of
+// corners left sharp because no maneuver fit within `cut_tol` of the track.
+size_t appendRoundedTrack(
+    F2CPath& path, const std::vector<F2CPoint>& poly, const F2CRobot& robot,
+    TurningBase& turn, double cut_tol, bool continuous,
+    const std::optional<double>& start_angle, const std::optional<double>& end_angle) {
+  const double radius = robot.getMinTurningRadius();
+  const double op_width = robot.getCovWidth();
+  const size_t n = poly.size();
+  const auto legAngle = [&poly](size_t a, size_t b) {
+      return (poly[b] - poly[a]).getAngleFromPoint();
+    };
+
+  // Deflection at each corner, which is what sizes the maneuver through it.
+  std::vector<double> sweeps(n, 0.0);
+  for (size_t k = 1; k + 1 < n; ++k) {
+    sweeps[k] = F2CPoint::getAngleDiffAbs(legAngle(k - 1, k), legAngle(k, k + 1));
+  }
+  if (start_angle && n > 1) {
+    sweeps.front() = F2CPoint::getAngleDiffAbs(*start_angle, legAngle(0, 1));
+  }
+  if (end_angle && n > 1) {
+    sweeps.back() = F2CPoint::getAngleDiffAbs(legAngle(n - 2, n - 1), *end_angle);
+  }
+
+  size_t sharp = 0;
+  F2CPoint cursor = poly.front();
+  size_t i = 0;
+  while (i < n) {
+    if (sweeps[i] < kMinSweep) {
+      addStraight(path, robot, cursor, poly[i]);
+      cursor = poly[i];
+      ++i;
+      continue;
+    }
+
+    // Corners too close together to round one at a time fold into one span.
+    size_t min_span = 1;
+    while (i + min_span < n && min_span < kMaxSpan &&
+        sweeps[i + min_span] >= kMinSweep &&
+        poly[i + min_span - 1].distance(poly[i + min_span]) <
+            2.0 * kRadiusMargin * radius) {
+      ++min_span;
+    }
+
+    bool rounded = false;
+    // Widening out from there, because a shorter span cuts less corner.
+    for (size_t span = min_span; span <= kMaxSpan && i + span <= n && !rounded; ++span) {
+      const size_t j = i + span - 1;  // last corner taken in one go
+      // At either end the pose is pinned to the swath, not slid along a leg.
+      const bool pin_in = (i == 0);
+      const bool pin_out = (j + 1 == n);
+      if (pin_out && !end_angle) {
+        break;
+      }
+      const F2CPoint& first = poly[i];
+      const F2CPoint& last_corner = poly[j];
+      const F2CPoint& after = pin_out ? poly[j] : poly[j + 1];
+
+      // Total turning sizes the maneuver's length; net deflection (which can
+      // be smaller, e.g. on an S where corners cancel) sizes its geometry.
+      double total = 0.0;
+      for (size_t k = i; k <= j; ++k) {
+        total += sweeps[k];
+      }
+      const double in_angle = pin_in ?
+          *start_angle : (first - cursor).getAngleFromPoint();
+      const double out_angle = pin_out ?
+          *end_angle : (after - last_corner).getAngleFromPoint();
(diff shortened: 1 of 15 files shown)


u33549:
Some before/after on the same scenarios, in case it helps reading the diff. Both columns run
the same code path on the same inputs — only this PR differs. `F2CRobot(2.1, 2.5)`,
`min_turning_radius` 0.4, `max_diff_curv` 2.0, `DubinsCurvesCC`. Blue is the track the route
hands to the planner, green-to-black is the path, shaded from the start of the run to its end.

### The connection is followed rather than cut across

A route whose connection detours over the top of two swaths. Before, the detour is skipped
and the path crosses at swath height; after, it follows the track and rounds the corners.

| before | after |
|---|---|
| 1_corners_smoothed before | 1_corners_smoothed after |

The corners are 0.4 m of radius in a 40 m plot, so here is the top-left one at 8 m across.
Before, there is no path in frame at all — the route never goes there.

| before | after |
|---|---|
| z1_corner_TL before | z1_corner_TL after |

### A curved connection stays curved

The connection here is a 40 m arc bulging 5 m off its chord. Before, the whole arc is
replaced by a single u-turn between the swath ends.

| before | after |
|---|---|
| 4_curved_track_kept before | 4_curved_track_kept after |

Left junction at 10 m across. This is also the limitation from the description: the entry
pose is pinned to the swath's end, so the turn has no approach room on that side and swings
a little wide before settling. Worst case in these scenarios, at a curvature rate well below
what a field machine uses.

| before | after |
|---|---|
| z4_junction_L before | z4_junction_L after |

### A corner too expensive to round is left square, not cut

Same field as the first pair, with an 8 m turning radius. There is room for the turn — the legs are 20 m and 30 m and the tangent needs 8 m — but at that radius the arc would cut almost 4 m off the corner, against a tolerance of half the operation width. So the corner stays square: a heading step the controller absorbs beats driving 4 m off the planned track.

| before | after |
|---|---|
| 3_unroundable_stays_sharp before | 3_unroundable_stays_sharp after |

### A decomposed field

A chaotic shape as a practical example

| before | after |
|---|---|
| 7_comb_field before | 7_comb_field after |

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+constexpr size_t kMaxSpan = 3;            // a spur, the corner it leads to, and a spur out
+
+// Below this deviation from the straight hop, a connection is driven direct.
+constexpr double kDirectHopDev = 0.25;
+// Points this close to the chord they'd round are dropped before rounding.
+constexpr double kSimplifyTol = 0.1;
+
+// cut_tol is the wider of these two -- see cutTolerance() for why.
+constexpr double kCutTolWidths = 0.5;
+constexpr double kCutTolRadii = 1.0;
```
Gonzalo-Mier: All these constants should go in the class Robot or the class PathPlanning, but in the .h file. Also, variables should be undercase with underscore as spaces. All these parameters should be also modificable (not constants and with getters and setters).

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+// width) still gets room for a fillet, but a wide-turning one doesn't get a
+// tolerance loose enough to cut corners it should instead leave sharp.
+double cutTolerance(const F2CRobot& robot) {
+  return std::max(
+      std::min(kCutTolRadii * robot.getMinTurningRadius(), robot.getCovWidth()),
+      kCutTolWidths * robot.getCovWidth());
+}
+
+// Point `dist` along the ray from `from` to `to`; `from` itself if they coincide.
+F2CPoint pointAlong(const F2CPoint& from, const F2CPoint& to, double dist) {
```
Gonzalo-Mier: Could this function be moved to the class F2CPoint?

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+    radius = std::max(radius, 1.0 / std::sqrt(rate * sweep));
+  }
+  return radius;
+}
+
+// How far rounding a corner may leave the track. Floored in turning radii,
+// capped at the operation width: a compact implement (radius close to its own
+// width) still gets room for a fillet, but a wide-turning one doesn't get a
+// tolerance loose enough to cut corners it should instead leave sharp.
+double cutTolerance(const F2CRobot& robot) {
```
Gonzalo-Mier: Can this function be moved to F2CRobot?

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+//
+//   - the clothoid ramps spend part of the deflection, so a turn that reaches
+//     full curvature sweeps wider than 1/kappa (getSmoothTurningRadius);
+//   - a shallow corner never reaches full curvature at all, peaking at
+//     sqrt(rate * sweep), so the gentler the corner the wider it sweeps.
+//
+// Taking the larger keeps the approach honest at both ends. Dropping the
+// deflection term lets a shallow corner be planned from a radius it never
+// drives, which starts the turn too late and overlaps the leg it came in on.
+double cornerRadius(const F2CRobot& robot, double sweep, bool continuous) {
```
Gonzalo-Mier: Can this be moved to F2CRobot?

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+constexpr double kDirectHopDev = 0.25;
+// Points this close to the chord they'd round are dropped before rounding.
+constexpr double kSimplifyTol = 0.1;
+
+// cut_tol is the wider of these two -- see cutTolerance() for why.
+constexpr double kCutTolWidths = 0.5;
+constexpr double kCutTolRadii = 1.0;
+
+// A leg runs along the connection's own track, driven at cruise speed.
+void addStraight(F2CPath& path, const F2CRobot& robot,
```
Gonzalo-Mier: Can this be added to F2CPath instead of here?

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+// Point `dist` along the ray from `from` to `to`; `from` itself if they coincide.
+F2CPoint pointAlong(const F2CPoint& from, const F2CPoint& to, double dist) {
+  if (from.distance(to) < 1e-9) {
+    return from;
+  }
+  return from.getPointFromAngle((to - from).getAngleFromPoint(), dist);
+}
+
+// Furthest any state of `arc` strays from the polyline it stands in for.
+double deviationFromTrack(const F2CPath& arc, const std::vector<F2CPoint>& track) {
```
Gonzalo-Mier: Could this be moved to F2CPath?

src/fields2cover/path_planning/path_planning.cpp
Gonzalo-Mier: Also, avoid using parameters that change the parameter itself.

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -4,11 +4,309 @@
...
+  }
+  return worst;
+}
+
+// Follow `poly`: straight runs as they are, corners through the turn planner.
+// `start_angle`/`end_angle` are the swath headings at either end, unset if
+// there's none -- the track's own first/last leg is often a spur off the
+// border graph the robot never drives (see PR-4, B.3). Returns the count of
+// corners left sharp because no maneuver fit within `cut_tol` of the track.
+size_t appendRoundedTrack(
```
Gonzalo-Mier: This function is 190 long. Could it be split into a few more functions so it is easier to follow the reasoning behind?

src/fields2cover/path_planning/path_planning.cpp
```diff
@@ -77,19 +375,45 @@ F2CPath PathPlanning::planPathForConnection(const F2CRobot& robot,
     const F2CMultiPoint& mp,
     const F2CPoint& p2, double ang2,
     TurningBase& turn) {
-  auto v_con = simplifyConnection(robot,
-      p1, ang1, mp, p2, ang2);
+  std::vector<F2CPoint> pts{p1};
+  for (size_t i = 0; i < mp.size(); ++i) {
+    pts.push_back(mp[i]);
+  }
+  pts.push_back(p2);
```
Gonzalo-Mier, line 338: This could be a new function in F2CMultiPoint called toVectorPoint.

src/fields2cover/path_planning/path_planning.cpp
Gonzalo-Mier: What does the output size_t means?

Gonzalo-Mier:
It looks super nice the changes! Thank you really much for all your efforts.

I asked to move a bit some functions, but the logic seems solid.

Thank you again @u33549

u33549:
@Gonzalo-Mier thank you, glad the approach reads well!

I can't see the comments about moving the functions on my side the diff shows none.

Happy to do the moves as soon as I can read them.

Gonzalo-Mier (changes requested):

Gonzalo-Mier:
> @Gonzalo-Mier thank you, glad the approach reads well!
> 
> I can't see the comments about moving the functions on my side the diff shows none.
> 
> Happy to do the moves as soon as I can read them.

Sorry, I forgot to send them

u33549:
Two structural questions before I do, since they change the shape of the diff:

**1. `getSmoothTurningRadius` and the radius helper.** The corner code picks a radius from `getSmoothTurningRadius` and from `sqrt(linear_curv_change * sweep)`. Moving that to `F2CRobot` means `getSmoothTurningRadius` has to move too, and it needs `end_of_clothoid` from `steering_functions` — which would pull that dependency into `types/`. It is also
public API as of #233, so moving it is a break unless `PathPlanning` keeps a forwarder. Would you rather it moved to `Robot` (with a forwarder left behind), or stayed in `PathPlanning`?

**2. Making the constants settable.** `PathPlanning`'s methods are all `static`, and `fields2cover.cpp` calls them as `PathPlanning::planPath(...)` in three places. Settable members mean the class gains state, so those calls become instance calls. Fine by me — I just want to confirm you are happy with `PathPlanning` becoming instantiable before I
change the facade.

My plan otherwise: the corner-cut tolerance onto `Robot` (it is a property of the machine and the implement), the rest of the tuning onto `PathPlanning`, `pointAlong` onto `Point`, the deviation and straight-leg helpers onto `Path`, `toVectorPoint` onto `MultiPoint`, and `appendRoundedTrack` split up. The `size_t` return is a sharp-corner count nothing reads — I'll drop it and make it `void`.

Gonzalo-Mier:
Hi @u33549,

Answering to your message:
1. It is okay to move that dependence to `F2CRobot`. You can also create a new function with that logic in `F2CRobot`, and keep the function `getSmoothTurningRadius` in `f2c::pp::PathPlanning` to keep the ABI.
2. Those parameters should be setteable, but shouldn't be placed in  `f2c::pp::PathPlanning`. Maybe a better place can be `F2CRobot`.

u33549:
Hi @Gonzalo-Mier,
`PathPlanning::getSmoothTurningRadius` stays as a forwarder, as you suggested.

Three places where I did not follow the letter of the review, each for a reason:

**`appendStraight` takes a speed, not a `Robot`.** `Robot.h` already includes `Path.h`, so `Path` cannot depend on `Robot`. It matches `appendSwath(swath, cruise_speed)`.

**The parameters that scale with the machine are stored in real units, not as the fractions they were.** `getMaxCornerCut()` is metres, defaulting to a value derived from the turning radius and the operation width; likewise `getDirectHopMaxDev`, `getUturnMaxHop`, `getTrackSimplifyTol`. An operator setting "at most 1.25 m of corner may be cut" seemed more useful than exposing a 0.5 coefficient. Unset values derive from the robot, following the `turn_vel_` precedent. The dimensionless ones (`getLegShare`, `getRadiusMargin`, `getMinBackoffRadii`, `getUturnReachRadii`, `getMaxCornerSpan`) are stored as they are.

**`toVectorPoint` went on `LineString` as well as `MultiPoint`.** The same loop appeared twice, two lines apart.

On the `size_t`: it was a count of corners left square, and nothing read it. Removed, so `appendRoundedTrack` returns `void`.

u33549:
One more: I'd resolved the thread about not mutating the passed-in `F2CPath` too early — moving the logic onto `Path::appendStraight` didn't help while a thin wrapper in this file still took `F2CPath&` and mutated it. Removed the wrapper in `d2805e3`; call sites use `path.appendStraight(...)` directly now.

Gonzalo-Mier:
One last thing before accepting it. You use comments with `\\\`, which are used by the documentation to explain what the functions do, but they read like independent comments. In this case you can make them just `\\`, or add a proper explanation to the function or the variable.

u33549:
I hope I've changed the comments correctly

Gonzalo-Mier:
Awesome changes and awesome PR! Thank you a lot again @u33549

u33549:
@Gonzalo-Mier You're welcome, it was a pleasure for me. I have a few more changes planned here, and I'm happy to work together on v3 once they're done.

u33549:
I’d also like to share what we’ve done on LinkedIn—if you’re okay with that, of course. If you approve, should I post about this change now, or wait until everything is finished and post everything at once?

Gonzalo-Mier:
Feel free to post everywhere what you do for open-source! no need to ask for permission or approval. It's your own work, so be proud wherever you want :smile:

PR #243, opened 2026-08-27
Carve a corridor between decomposed cells, and fix three silent failures

u33549:
Decomposing a field and then covering it currently forces a choice between two bad
options, and the tutorial itself takes the worse one. This PR adds the missing operation
and fixes two crashes/silent skips found while building it.

## The problem, in the project's own words

`docs/source/tutorials/decomposition.rst` explains the decomposition + route planner
workflow like this:

> The decomposition workflow doesn't work directly with the route planner, because the
> middle headland ring of each decomposed cell are not connected between them.
>
> To solve this, we use **a simple trick**. First, we create the middle headland ring with
> the headland generator. Second, we decompose the inner part of the middle headland ring.
> Third, we generate the mainland using the headland generator **again**.

That third step is the problem. `ConstHL::generateHeadlands` shrinks *every* border, so
running it a second time takes another headland off the outer boundary that already has
one. The cells do get separated — but the whole field also loses a second ring it never
needed to lose.

What is missing is an operation that opens a corridor **only where two cells meet**.

## What this adds

```cpp
/// Carve a corridor of \p width where cells of this set border each other.
/// Edges facing the outer boundary or a void are left untouched.
F2CCells carveSharedBorders(double width) const;
```

For each border edge, the part a neighbour actually touches is intersected and cut. Two
details matter:

- **Only the touching part of an edge is cut.** An edge that is half shared gets half a
  corridor. Testing whole edges (endpoints + midpoint) misses that case entirely.
- **The whole corridor comes out of the cell with the shorter border.** Taking half from
  each side notches both cells, and a notched cell costs the swath generator a pass — the
  narrow arm gives way, the large band keeps its rectangle.

The decomposition tutorial now uses it instead of the second headland.

## Numbers

Full pipeline — headland → decompose → swaths → route → path (`DubinsCurvesCC`),
`F2CRobot(2.1, 2.5)`, `minTurningRadius = 0.4`, `maxDiffCurv = 2.0`.

**The outer headland is equal on both sides (6.00 m, measured, not assumed)**, so the only
difference left is the method. `A` is today's double-headland flow, `B` is the corridor.

| field | | outer | gap | area | swaths | swath m | connect m | path m | efficiency |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| **comb** | A | 6.00 | 6.00 | 6750.0 | 28 | 2625.0 | 427.6 | 3052.6 | 86.0% |
| | B | 6.00 | 2.50 | **6939.0** | 28 | 2698.5 | **402.2** | 3100.7 | **87.0%** |
| | *diff* | *0.00* | | ***+3%*** | *0* | *+73.5* | *−25.4* | *+48.1* | *+1.0pp* |
| **L-shape** | A | 6.00 | 6.00 | 1136.0 | 6 | 426.0 | 30.0 | 456.0 | 93.4% |
| | B | 6.00 | 2.50 | **1164.0** | 6 | 436.5 | **22.9** | 459.4 | **95.0%** |
| | *diff* | *0.00* | | ***+2%*** | *0* | *+10.5* | *−7.1* | *+3.4* | *+1.6pp* |
| **cross** | A | 6.00 | 6.00 | 4928.0 | 33 | 1936.0 | 324.5 | 2260.5 | 85.6% |
| | B | 6.00 | 2.50 | **5124.0** | 33 | 2013.0 | **250.8** | 2263.8 | **88.9%** |
| | *diff* | *0.00* | | ***+4%*** | *0* | *+77.0* | *−73.7* | *+3.3* | *+3.3pp* |
| **holed** | A | 6.00 | 6.00 | 6078.0 | 36 | 2388.0 | 291.6 | 2679.6 | 89.1% |
| | B | 6.00 | 2.50 | **6540.0** | 42 | 2586.0 | **245.4** | 2831.4 | **91.3%** |
| | *diff* | *0.00* | | ***+8%*** | *+6* | *+198.0* | *−46.3* | *+151.7* | *+2.2pp* |

*efficiency = swath length / total path length — how much of the driving does work.*
*gap = measured clearance between neighbouring cells; outer = measured clearance from the
field boundary.*

Across four fields: **workable area +2…8%**, **connection distance −7…−74 m**,
**efficiency +1.0…+3.3pp**. The gain scales with headland width — measured on `comb`, an
outer headland of 2.5 m gives 0%, 6.0 m gives 3%, 10.0 m gives 6%.

### Pictures

Each pair is drawn in the same window, so the two can be laid over each other.

| field | today (double headland) | with corridor |
|---|---|---|
| **comb** |  |  |
| **L-shape** |  |  |
| **cross** |  |  |
| **holed** |  |  |

In `cross` the middle band keeps its rectangle instead of being notched where the arms
meet it. In `holed` the headland around the obstacle is preserved and corridors open only
on the cell borders.

## Three bugs found while building it

All three are reachable without the new API, and each one fails silently or crashes.

**1. `Cells` accessors dereference a null polygon.**

```cpp
F2CCells cells;
cells.addGeometry(F2CCell());   // one empty polygon
cells.getCellBorder(0);         // segfault
```

An empty polygon has no exterior ring, and `getGeometryRef()` returns a null entry for an
out-of-range index — which `OGRMultiPolygon::getGeometryRef` then dereferences. Neither
was checked. `getCellBorder`, `getInteriorRing` and `addRing` now check the index before
asking OGR, and throw `std::out_of_range` like `getGeometry` already does. The `Geometry`
constructors also accept a null pointer as "empty geometry" instead of calling
`clone()` on it. Confirmed under ASan.

**2. `generateBestSwaths` can return an angle that covers nothing.**

The objectives score an angle from the cell border alone, without checking whether any
swath is produced. On a cell left with a hairline spur by a boolean operation:

```
angle=0    swaths= 0   cost=17.50   <- cheapest, and the one chosen
angle=90   swaths=15   cost=22.50
angle=135  swaths=19   cost=20.51
```

The cell was skipped without a word. `generateBestSwaths` now falls back to an angle that
actually covers the cell. The check sits where the swaths are already generated, so the
common path costs nothing measurable (0.7414s vs 0.7366s over 480 calls).

**3. `NSwathModified::computeCost` reads the wrong neighbouring point.**

```cpp
auto p_1 = ring.getGeometry((i - 1) % ring.size());
```

`i` is a `size_t`, so at `i = 0` the subtraction wraps to `SIZE_MAX` and the modulo lands
on an arbitrary vertex rather than the previous one — index `0` for a 5-point ring, `3`
for a 6-point ring, `1` for a 7-point ring. The first edge is therefore weighed against an
unrelated point.

It shows up as the cost of a polygon depending on which vertex its ring happens to start
from: for one 5-sided polygon written five different ways, the cost ranged from 10.5 to
15.0. Fixed by `(i + ring.size() - 1) % ring.size()`; no existing test moved.


Gonzalo-Mier:
Hi @u33549, please, tell me when this is ready to review/merge

u33549:
Hi @Gonzalo-Mier, i think it's ready but I was quite confused while developing, there may be parts I missed.

Gonzalo-Mier:
Hi @u33549, Wow. I'm seriously out of words. Awesome job!!!

Just a few things. 

1. I feel like this should be its own class of headland decomposition instead of a method of constant width. Part of the purpose of F2C is being able to compare algorithms easily, and adding this awesome method as part of a (worst) algorithm could lose the objective of this comparison.

2. I'm curious to see how does it behaves when the frontier of the more than 2 cells collides in a point. Could you make a 5 star polygon, so it has 5 cells that shares frontiers equally?

Thank you again @u33549

u33549:
With the second one you mentioned, you really did find a bug, but I have a solution in mind—I'll try it out and send it to you.

u33549:
## What the five-cell case showed

Five equal slices meeting at one point, corridor width 4.0. Every slice should give up the
same amount. It did not:

| | input | current | fixed |
|---|---|---|---|
| |  |  |  |
| slice areas | 1498.9<br>1498.9<br>1498.9<br>1498.9<br>1498.9 | 1125.5<br>1303.8<br>**1498.9**<br>1303.8<br>1303.8 | 1305.5<br>1305.5<br>1305.5<br>1305.5<br>1305.5 |
| carved away | — | 958.6 | 966.8 |
| corridors opened | — | **3 of 5** | **5 of 5** |

In the middle picture two of the five borders get no corridor at all, and one slice keeps
its full area — the robot cannot cross there.

## Why

The rule was "the cell with the shorter border gives up the corridor", so that a narrow arm
yields and its larger neighbour keeps its rectangle. In the star all five perimeters are
equal in theory, but not in floating point:

```
h0 = 162.45040684215815
h1 = 162.45040684215820
h2 = 162.45040684215829   <- largest by 1.4e-13
h3 = 162.45040684215820
h4 = 162.45040684215815
```

A strict `>` let that 1e-13 of noise decide the winner. `h2` came out "largest" against
every neighbour, so it never gave anything, and the borders around it stayed closed.

## Fix

Compare the perimeters with a tolerance, and when the cells are the same size let them
split the corridor evenly:

```cpp
const bool same_size =
    std::abs(perimeter - perimeter_k) <= kSameSize * perimeter;
if (!same_size && perimeter > perimeter_k) {
  continue;                                   // the smaller neighbour gives
}
const double share = same_size ? 0.5 : 1.0;   // equals split it
```

A real size difference (the cross field: arm ~145 vs band ~290) still yields one-sided, so
the band keeps its rectangle. Only ties now split.

One existing expectation moved with it: two equal 50x100 cells used to end up 48x100 and
50x100, and are now 49x100 each. Same 200 m² removed, shared evenly.

Gonzalo-Mier:
Thank you for the PR, @u33549! We have a new planner

PR #247, opened 2026-08-29
Backport the required-width headland generator from v3.0

u33549:
This is `f2c::hg::ReqHL` from the `v3.0` branch, brought to the 2.x line together with the ring API it needs.

## What it does, briefly

A headland exists so the machine can turn, and how much room a turn needs depends on the angle between the swaths and the border they meet. A border the swaths run *along* is only driven into and left again; a border they *end on* takes a whole turn. `ConstHL` cannot tell the two apart, so it charges every border the second figure:

```
(1 + |sin(track_ang - border_ang)|) * getMinTurningRadius() + 0.5 * getWidth()
```

which runs from `R + W/2` to `2R + W/2 == maxHLWidthRequired`. On a 100 m square with a 3 m machine and a 5 m turning radius, `ConstHL` takes 11.5 m off all four borders where two of them need 6.5 m.

## What came across unchanged

Taken verbatim from `origin/v3.0` (the same files in `pre-v3.0` are byte-identical, so there was no version to choose between):

- `include/fields2cover/headland_generator/required_headland.h`
- `src/fields2cover/headland_generator/required_headland.cpp`
- the `HeadlandGeneratorBase` additions: `generateHeadlands(cell, robot, track_ang)`, `generateHeadlands(cells, robot, track_angs)`, `maxHLWidthRequired`, with their default bodies
- from `LinearRing`: `getParallelLine` (both overloads), `bufferOutwards`, `bufferInwards`, `filterSelfIntersections`, `removePoint`, `getSegment`, `getLastSegment`, `segmentLength`, `segmentAng`
- `Geometry::contains`
- `%rename(HG_Req_gen)`, the name `v3.0`'s own `.i` file gives the class

The mainland is built the way `v3.0` does it: take the widest headland off once, then give each border back what it does not need, so no border is cut twice; interior rings the same way with the sign reversed, so an obstacle grows into the field.

## What we changed, and why

**1. Constness.** `v3.0` declares the whole headland interface `const`; nothing in the 2.x interface is. The two new overloads are therefore non-const here. Making them `const` would have meant changing `ConstHL`'s signatures and breaking every out-of-tree generator in the 2.x line.

**2. `generateHeadlandSwaths` returns `std::vector<F2CCells>`,** this line's signature, instead of `v3.0`'s `std::vector<F2CMultiLineString>` — the base class differs between the lines. The body is `ConstHL`'s, as it is in `v3.0`.

**3. The clean-up buffer now does what it says.** In both `v3.0` and `pre-v3.0`:

```cpp
new_cell = new_cell.buffer(new_cell, 1e-5).buffer(new_cell, -1e-5);
```

`Cell::buffer` is static, so calling it through an instance ignores the instance: the second call re-buffers the *original* `new_cell` and the dilated result of the first is discarded. What reads as a closing — dilate, then erode, to heal the hairline spurs that per-segment offsetting leaves — is in fact a plain 1e-5 erosion, applied once per ring and once per cell.

Three builds compared over 200 random fields on area, cell count and ring count:

| | area differs | shape differs |
|---|---|---|
| as written vs. a real closing | 200 / 200, up to 0.008 m² | never |
| a real closing vs. no clean-up at all | 1 / 200, 7e-6 m² | never |

The line as written never cleans anything, it only shrinks. Here it reads `F2CCell::buffer(F2CCell::buffer(new_cell, 1e-5), -1e-5)`, and the same for the `F2CCells` line below it. **These two lines are still as they were in `v3.0`.**

**4. `%feature("notabstract")` for the concrete generators.** `ConstHL` and `ReqHL` pull the base overloads in with `using HeadlandGeneratorBase::generateHeadlands;` — needed in C++, since declaring a member of that name otherwise hides every base overload, and present in `v3.0` too. SWIG reads that declaration as an inherited pure virtual and marks the class abstract, so `f2c.HG_Const_gen()` raises `AttributeError: No constructor defined - class is abstract`. The `.i` file now says otherwise. **This affects `ConstHL` in `v3.0` as well** — it looks like the bindings were never exercised there after the `using` was added.

**5. Dropped:** the `#include "fields2cover/utils/visualizer.h" // TODO: Remove this`, and the `// If d>0, the result is at the right side of the curve.` comment on `getParallelLine` — the offset goes to the *left* of the segment direction (`ang + 0.5 * M_PI`), which is what the tests measure.

**6. Tests.** This is the largest thing we added. `v3.0`'s `req_hl_test.cpp` has one test that asserts nothing at all (it builds a field and plots it, commented out) and three that exercise only the constant-width methods; `bufferInwards` has a single test and the rest of the ring API has none. 320 → 337 C++ tests and 106 → 111 Python tests here, all green, no existing test touched. Each formula was mutation-tested: reverting the angle term drops 5 tests, the interior-ring sign drops the obstacle test, the `2 *` in `maxHLWidthRequired` and the default dispatch each drop their own.

They cover the width formula at 0°, 90°, 45° and its bound; one distance per segment rather than per point; the exact mainland rectangle and its transpose; an obstacle inside the field; one track angle per cell and the documented fallback without them; and the ring API on its own — parallel lines with a uniform and a per-segment offset, buffering in both directions from either winding, self-intersection filtering on a bow tie and on a doubled-back spur.

**7. Tutorial and docs.** `v3.0` does not document `ReqHL` — its `headland_generator.rst` is identical to this line's. Section 3.2 is new, in both languages, with a figure. Plus a CHANGELOG entry.

**8. One adjacent one-liner** (`64612cc`): section 3.1 said *"On orange it is `cells` and in green it is `no_hl`"*, which was true until the figures were regenerated with another plotting backend in `c760058` — they have been blue and orange since. Easy to drop from the branch if you would rather keep this PR to one subject.

We did not touch the behaviour anywhere else: the fallback when the number of track angles does not match the number of cells still returns a constant headland rather than throwing, as in `v3.0`. We only made it explicit in a test.

## Not in this PR

The headland *swaths* are still constant-width: `generateHeadlandSwaths` buffers by a fixed distance, as it does in `v3.0`, where the class carries a `// TODO: Return the headland swaths`. Only the mainland boundary varies with the border angle here.


Gonzalo-Mier:
Thank you for the port and the PR, @u33549

PR #248, opened 2026-08-30
Report which cell gives each corridor, and let callers ask for a fair split

u33549:
`CorridorHL::generateHeadlands` decides, for every pair of cells that share a border, which one gives up the corridor and how much — but that decision was made and used in the same breath, inside the carving loop. Nothing outside the carved geometry could see it. This PR pulls the rule out into its own method, fixes a real bug the pull surfaced, and adds a second rule to compare the first one against.

## `corridorShares`, the rule on its own

```cpp
namespace f2c::hg {
struct CorridorShare {
  size_t cell_i, cell_k;
  double perimeter_i, perimeter_k;
  bool same_size;
  double share;           // part of the corridor cell_i gives up: 0, 0.5 or 1
  double shared_length;
  F2CMultiLineString shared_border;
};

std::vector<CorridorShare> CorridorHL::corridorShares(
    const F2CCells& field, CorridorShareMode mode = CorridorShareMode::ASYMMETRIC) const;
}
```

`generateHeadlands` now calls this and applies what it returns — same output as before, just no longer the only way to find out. A caller that needs the decision itself (a report, a test, a figure) doesn't have to re-derive the rule from the carved shape and risk it drifting from what the generator actually does.

## A bug the pull surfaced

Finding the shared border buffers the neighbour cell by a tolerance and intersects it with each edge. That tolerance turns a single shared *corner* into a few millimetres of "border" on every edge that reaches it. It's invisible with two cells, but five cells meeting at one point were all reported as neighbours of every other, and the corridor carved a disc out of the middle instead of five separate cuts. Fixed by giving the length filter its own, much looser threshold (`kMinBorder = 1e-2`) than the buffer tolerance (`kTol = 1e-3`) — a border under a centimetre is a corner, not a border. Covered by `sharesTreatACornerAsNoBorder`, a five-slice pie test.

## `CorridorShareMode`: a second rule to compare against

`corridorShares` always gave the whole corridor to the smaller cell (by perimeter), splitting evenly only when the two were within tolerance of the same size. There was no way to ask for the even split on purpose, which is exactly what's needed to measure the asymmetric rule's effect — the two only differ in *which* cell gives up how much, not in how much corridor opens overall.

```cpp
enum class CorridorShareMode { ASYMMETRIC, SYMMETRIC };

std::vector<CorridorShare> corridorShares(const F2CCells&, CorridorShareMode) const;
void setShareMode(CorridorShareMode);        // what generateHeadlands() uses
CorridorShareMode getShareMode() const;
```

`ASYMMETRIC` stays the default, so nothing that already calls `generateHeadlands` or the no-mode `corridorShares` overload changes behaviour.

## Tutorial

Tutorial 7.2 had grown a `corridorShares` demo into the same figure as the route, at the same zoom as every other headland tutorial — three different concepts drawn on top of each other, indistinguishable at a glance. Removed. In its place, Tutorial 3.3 gives `CorridorHL` its own section: one L-shaped field, decomposed once, headlanded three different ways (double headland, `ASYMMETRIC` corridor, `SYMMETRIC` corridor), each with its own route and path (`DubinsCurvesCC`) so the cost difference is visible, not just the leftover area.


src/fields2cover/headland_generator/corridor_headland.cpp
```diff
@@ -11,35 +11,45 @@
...
-  constexpr double kSameSize = 1e-9;
-  // Cut along the part of each border edge that a neighbour actually touches.
+namespace {
+constexpr double kTol = 1e-3;
+constexpr double kSpur = 1e-9;
+constexpr double kSameSize = 1e-9;
+// Buffering the neighbour by kTol turns a shared corner into a piece a few
+// millimetres long on each edge that reaches it. A border that short is a
+// corner, and a corner carries no corridor.
+constexpr double kMinBorder = 1e-2;
```
Gonzalo-Mier: Could you move these parameters as private to the header file? They should be visible in the header class directly

u33549:
CorridorHL currently creates corridors that are wider than they should be; this will be fixed in the next update.

src/fields2cover/headland_generator/corridor_headland.cpp
u33549: Done

u33549:
@Gonzalo-Mier I think you forgot to close this PR, so I wanted to remind you.

Gonzalo-Mier:
Thank you for the notice and the PR @u33549

PR #249, opened 2026-09-01
Fix Cells::splitByLine throwing on a pinched split

u33549:
A split can leave a piece touching itself at a single point (a pinch). `Cells::splitByLine` reinflated each split piece through `Cell::buffer`, which only accepts a single polygon back; a positive buffer on a pinched piece can separate it into two, and `Cell(const OGRGeometry*)` throws `std::invalid_argument` on the resulting `MultiPolygon` instead of decomposing the field.

## Reproduction

`data/ee_field_130.wkt` is a real field. Decomposing it with `BoustrophedonDecomp` throws at 12 of 180 swept split angles (58, 60, 131, 142, 144, 147, 148, 152, 155, 160, 162, 168 degrees). Added as `fields2cover_decomp_boustrophedon.doesNotThrowOnAPinchedSplit`, using angle 58.

## The fix

```cpp
Cells Cells::splitByLine(const LineString& line) const {
  return this->difference(this->buffer(line, 1e-8));
}
```

The reinflate step turns out to have never run:

```cpp
for (auto&& c : cells) {
  c = Cell::buffer(c, 1e-8 * 0.5);
}
```

`c` is the range-for's by-value loop variable; assigning to it never writes back into `cells`. Verified directly — the container's WKT is byte-identical before and after this loop runs. So removing it fixes the crash and changes nothing for pieces that were never pinched, since the step already had no effect on them. The existing `fields2cover_decomp_boustrophedon.decompose` test (an unrelated shape, no pinch involved) passes unchanged.


tests/cpp/decomposition/boustrophedon_decomp_test.cpp
```diff
@@ -25,3 +29,34 @@ TEST(fields2cover_decomp_boustrophedon, decompose) {
...
+  std::stringstream ss;
+  ss << f.rdbuf();
+  std::string wkt = ss.str();
+  OGRGeometry* geom = nullptr;
+  const char* c = wkt.c_str();
+  OGRGeometryFactory::createFromWkt(&c, nullptr, &geom);
+  F2CCell cell(geom);
+  OGRGeometryFactory::destroyGeometry(geom);
+  return cell;
+}
```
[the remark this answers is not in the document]
u33549: Fixed: switched to importFromWkt (no raw pointer, no null-deref risk), and loadWktCell now fails the test explicitly (ADD_FAILURE on a missing file, ASSERT_GT(area, 0) on a failed parse) instead of silently proceeding on an empty field. Mutation-tested by breaking the fixture path and confirming a clean assertion failure, not a crash.

include/fields2cover/types/Geometries_impl.hpp
```diff
@@ -41,6 +41,9 @@ Geometries<SAMETYPE, T, R, CHILDRENTYPE>::Iterator::Private::~Private() {}
 
 template <class SAMETYPE, class T, OGRwkbGeometryType R, class CHILDRENTYPE>
 void Geometries<SAMETYPE, T, R, CHILDRENTYPE>::Iterator::update() {
+  // T not listed here means `for (auto&& c : geoms) { c = ...; }` silently
+  // writes into this iterator's own copy and never reaches `geoms` --
+  // no error, no warning, just a no-op. OGRMultiPolygon (Cells) is one.
```
Gonzalo-Mier: Why is this comment added?

src/fields2cover/types/Cells.cpp
```diff
@@ -192,11 +192,7 @@ Cells Cells::unionCascaded() const {
 }
 
 Cells Cells::splitByLine(const LineString& line) const {
-  Cells cells = this->difference(this->buffer(line, 1e-8));
-  for (auto&& c : cells) {
-    c = Cell::buffer(c, 1e-8 * 0.5);
-  }
```
Gonzalo-Mier, line 198: This reinflation is necessary to keep the cells the same size at the beginning and at the end. If the problems is that cells si not properly modified, the fix is to do a for loop to write on it, not to remove it.

include/fields2cover/types/Geometries_impl.hpp
u33549: We take some notes inline while working with AI; I forgot to delete them.

src/fields2cover/types/Cells.cpp
u33549: You're right, and I was wrong to remove it. Restored the reinflate through `Cells::buffer` instead of `Cell::buffer`, so a piece coming out pinched into two no longer crashes — it collects into a fresh `Cells` rather than being written back to a single slot via `setGeometry`.

While fixing this I found a second, related issue: reinflating after every individual line in a `MultiLineString` split let GEOS collapse the next line's cut into a no-op (the two operations are close enough in scale that GEOS's overlay treats the second cut as already accounted for). So `splitByLine(MultiLineString)` now buffers every split line together and cuts in a single pass instead of one line at a time — this also fixes a latent bug where only the first line's cut actually took effect.

`tests/cpp/types/Cells_test.cpp`'s `splitByLine` test needed a small update: cutting all lines at once no longer guarantees the same piece ordering as cutting sequentially, so it now checks the four resulting areas as a sorted set instead of by index.

Gonzalo-Mier:
Nice, thanks @u33549 for finding this bug!

PR #252, opened 2026-09-18
Hand a graph's nodes back in id order

u33549:
`Graph2D::getNodes()` walked the `unordered_map` the nodes are kept in, so the
vector came back in hash order: not the order the ids were handed out in, and not
the same order on another build. The ids `getEdges()` reports index exactly that
vector, so a caller resolving an edge through it joined the wrong pair of points --
silently, and differently on another machine.

Measured on a visibility graph over a real field: of 1911 edges, 933 appeared to
run through the crop; resolved in id order, 2.

Ids are dense over `0..numNodes()-1` because they are handed out as the node count
at insertion and nodes are never erased, so the lookup cannot throw.

Adds a test that the ids `getEdges()` reports index the vector `getNodes()` returns.


Gonzalo-Mier:
Thank you for finding all this issues @u33549

PR #253, opened 2026-09-19
Drop border detail the machine cannot act on before decomposing

u33549:
A decomposition splits a field wherever its border turns back on itself. A
digitised border does that on details of a few centimetres -- far below
anything the machine can act on -- so the field comes back cut into pieces that
exist only in the survey, each one paying for a headland it does not need.

`f2c::decomp::simplifyForDecomposition(cells, robot)` removes border detail
below 0.6 of the robot's coverage width before the field is decomposed.
`ee_field_130` goes from sixteen cells to twelve.

**Measured over 305 real fields x two robot profiles** (2.5 m and 6 m coverage,
610 field/robot pairs):

- 2425 -> 1787 cells out of the decomposition (-26%)
- no piece of any field lands outside the field it came from, at any threshold
  up to 1.6 of the coverage width

**On a 50-field subset with the full coverage metric** (59 pairs, mainland minus
the area a turn overruns): the constant-width headland gains **+15,646 m²** and
the classic `2R + w/2` headland **+9,364 m²**. The gain is largest for the
simplest method, which is what you would expect if it is removing work nobody
should have been doing.

The threshold is measured rather than assumed. Below it noise survives; well
above it the splits the shape genuinely asks for start merging -- at 1.2 of the
coverage width a nine-cell field collapses to one.

`decompose()` is untouched. It still returns cells whose area equals its input,
so nothing changes for callers that do not ask for the simplification.

**One deliberate limitation:** the result is not clipped to the input.
Simplifying moves the border both ways, so a point can land outside; the
function expects ground that is already inside the field -- a mainland from a
headland generator, not the raw boundary -- and says so in its doc comment.
Clipping would guarantee containment but costs most of the benefit: on the same
50 fields the cell reduction drops from 33% to 13% and the area loss grows
tenfold.


src/fields2cover/decomposition/decomposition_base.cpp
```diff
@@ -25,5 +25,21 @@ F2CCells DecompositionBase::merge(
...
 
+namespace {
+// Part of the coverage width a border detail has to exceed to survive.
+//
+// Measured over 305 real fields and two robots: at this share the
+// decomposition returns 30% fewer cells, no piece of any field lands outside
+// the field it came from, and the area moves by less than a tenth of a
+// percent. Taking less leaves noise behind; taking much more starts merging
+// splits the shape of the field genuinely asks for.
+constexpr double kBorderDetailShare = 0.6;
```
Gonzalo-Mier: This value should be part of the class or argument of the function used. You can add it as last argument and give it a default value if you want, but never here.

Gonzalo-Mier (changes requested):

src/fields2cover/decomposition/decomposition_base.cpp
u33549: Actually, I chose this value based on the results I obtained by calculating it for 300 scenarios using approximately 20 different values (for a total of 6,000 scenarios), but it makes sense to make it optional.

Gonzalo-Mier: I understand the way you picked the value, but I dont like having random constants in the cpp files.

u33549: Done in c5ad294 — the constant is gone from the .cpp. It is now the last argument, `double detail_share = 0.6`, and the measurement behind the default moved into the parameter's doc comment in the header.

Gonzalo-Mier:
Thank you @u33549!!

PR #254, opened 2026-09-22
Add a common interface for the route generators

u33549:
`f2c::rp` has five route planners in two families: the swath orderers — `BoustrophedonOrder`, `SnakeOrder`, `SpiralOrder` and `CustomOrder`, all under `SingleCellSwathsOrderBase` — and the TSP planner, `RoutePlannerBase`. Since #251 all five do the same job, `(F2CCells, F2CSwathsByCells) → F2CRoute`, with the same call. But the two families don't know about each other: no common base, no common virtual method. A caller that lets the user choose the planner has to keep a code path per family:

```cpp
F2CRoute route;
if (use_tsp) {
  route = tsp.genRoute(cells, swaths);
} else {
  route = orderer->genRoute(cells, swaths);   // SingleCellSwathsOrderBase*
}
```

This adds the base both families implement, so the choice is made once and everything after it holds one type.

## The interface

```cpp
namespace f2c::rp {

/// Common interface of everything that turns swaths into a route.
class RouteGeneratorBase {
 public:
  virtual F2CRoute genRoute(
      const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
      double d_tol = 1e-4) const = 0;

  virtual ~RouteGeneratorBase() = default;
};

}  // namespace f2c::rp
```

It depends on `fields2cover/types.h` only, so nothing from OR-Tools reaches it. The signature is the one `SingleCellSwathsOrderBase::genRoute` got in #251, which is also the one `v3.0` has there — the base carries over to `v3` without a signature change. `v3.0` has no common base either: it grew `genRoute` to eight overloads on the TSP side and four on the orderer side, still unrelated.

## How each family joins it

**`SingleCellSwathsOrderBase`: one word.** Its `genRoute` already has exactly this signature, so it gains a base class and `override`. The body is untouched, and none of the four orderers overrides `genRoute`, so none of them changes.

**`RoutePlannerBase`: a second overload.** Its `genRoute` takes the optimizer settings and is not `const`:

```cpp
F2CRoute genRoute(const F2CCells&, const F2CSwathsByCells&,
    bool show_log = false, double d_tol = 1e-4, bool redirect_swaths = true,
    long int time_limit_seconds = 1, bool search_for_optimum = false);
```

That one stays as it is. The interface is answered by a new `const` overload with the default settings, and both delegate to one private `const` body — the current body, moved unchanged, since it only calls `const` methods and writes no members:

```cpp
F2CRoute genRoute(const F2CCells& cells, const F2CSwathsByCells& swaths_by_cells,
    double d_tol = 1e-4) const override;
```

Making the existing overload `const` instead looks simpler and is wrong: with both `const`, `genRoute(cells, swaths)` fits either one equally well and stops compiling — and `src/fields2cover.cpp:118` makes exactly that call. Left non-const, the existing overload wins that tie on a non-const planner, so the two-argument call keeps going where it went.

## Source compatibility: one call shape changes

Each shape compiled against this branch, and the overload read back from the object file's symbols:

| call on a non-const `RoutePlannerBase` | before | GCC 11 | Clang 14 |
|---|---|---|---|
| `genRoute(cells, swaths)` | settings overload | settings overload | settings overload |
| `genRoute(cells, swaths, true)` | settings overload | settings overload | settings overload |
| `genRoute(cells, swaths, false, 1e-3)` | settings overload | settings overload | settings overload |
| **`genRoute(cells, swaths, 1e-3)`** | settings overload, **`show_log = true`** | **new overload, `d_tol = 1e-3`** | **ambiguous — does not compile** |

The last shape is the exception, and it deserves a word. Before this PR it compiled and did something nobody writing it meant: the double went into `show_log`, and `d_tol` stayed at its default. GCC now sends it to the new overload, where the value lands in `d_tol` — the call starts doing what it reads as. Clang calls it ambiguous: the settings overload is the better match for the object, the new one for the third argument, and neither wins both.

Nothing in the repository makes that call — every `genRoute` in the library, the tutorials, the C++ and Python tests passes either two arguments, or an explicit `bool` third. A caller outside that does write it gets a compile error on Clang, pointing at the line where it was already passing a tolerance as a flag. If you'd rather it kept compiling everywhere, the way out is giving the new overload a name of its own instead of overloading `genRoute`, at the cost of the interface method no longer being called `genRoute` on the TSP planner.

Calling `genRoute` with the settings on a `const RoutePlannerBase` does not compile — but it didn't before either, since the settings overload was never `const`.

## Python

The base is exposed as `RP_Route_generator_base_class`, following the naming the other bases use. `route_generator_base_test.py` checks that all five planners are instances of it on the Python side too — not just classes that happen to have a `genRoute` — and routes a field with each.

## Tests

`route_generator_base_test.cpp`:

- `everyPlannerAnswersTheSameCall` — all five planners, held as `unique_ptr<RouteGeneratorBase>`, route the same field, with a connection for every swath
- `orderersRouteTheSameThroughTheBase` — an orderer gives the same route called directly and through the base
- `tspRoutesTheSameThroughTheBase` — the same for the TSP planner
- `deletingThroughTheBaseIsSafe` — the virtual destructor
- `constPlannersRoute` — the interface method is callable on a `const` orderer and a `const` TSP planner, and gives the same route as on a non-const one

`route_planner_base_test.cpp` gains `genRouteKeepsItsCallShapes`: the two-, five- and seven-argument calls still compile, reach the settings overload and give the same route.


docs/source/tutorials/route_planning.rst
Gonzalo-Mier, line 1: Tutorial should be generic for newcomers. This addition reflects more what have changed from what was before. I believe the  complete tutorial could be rewritten to present everything together in a more friendly way :)

Gonzalo-Mier (commented):
My proposal will be to remove the tutorial changes and add an issue for v3.0.0 updating the tutorials

u33549:
It makes sense. I will remove it at the appropriate time.

u33549:
Removed in 3ea81fb — the tutorial is back to what it is on `develop`.

You know the tutorials much better than I do, so I'd leave the v3.0.0 issue to you, to frame it the way the rewrite should go. Two things I ran into that might be worth a line in it, if useful:

- The swath orderers and the TSP planner now answer the same `genRoute(cells, swaths_by_cells)` call through `RouteGeneratorBase`, so a route planning tutorial could choose the planner once rather than walk through two separate flows.
- `5_route_planning.cpp` orders the swaths of a single cell (`no_hl.getGeometry(0)`), which sidesteps the multi-cell case the orderers handle since #251.

Gonzalo-Mier:
Thank you @u33549. I will take that into account when rewritting the tutorials. Thank you for the PR again!

u33549:
@Gonzalo-Mier, to avoid breaking the API, we’re mostly just making some additions and not removing much. Sometimes I feel like the codebase is getting too bloated. Do you have any plans regarding this?

Gonzalo-Mier:
> @Gonzalo-Mier, to avoid breaking the API, we’re mostly just making some additions and not removing much. Sometimes I feel like the codebase is getting too bloated. Do you have any plans regarding this?

That's a really good point. What do you want to remove? We can of course simplify things. Preferably, these changes should go to v3.0.0, as a way to tell the user there are major changes.

Just as a note, the objective of the library is help other scientists to compare CPP algorithms, so we should try to avoid removing final algorithms just because they are not optimal.

u33549:
I'm not sure right now, but I'm certain there must be a way to remove certain features without removing the algorithms.

Gonzalo-Mier:
I'm super open to this. Tell me if you look at it :)

PR #255, opened 2026-09-22
Drive the cells in the order they are worth reaching

u33549:
`SingleCellSwathsOrderBase::genRoute` — the route BOUSTROPHEDON, SNAKE, SPIRAL and CUSTOM produce since #251 — drives the cells in the order the decomposition happens to list them, and enters each one from the same end. That order says nothing about the field: the route crosses it to reach a cell it could have taken on the way, and enters cells against the direction it arrived from.

## What changes

**The next cell is the cheapest one to reach** from where the last one was left: the distance through the headland graph — so a cell on the other side of a wall is not taken for a near one — plus the turn needed to line up on it. Where the graph cannot connect two cells, as between two separate fields, it falls back to the straight-line distance.

**A cell is driven in reverse** when its far end is the nearer one.

**The start is given, not searched for.** The route starts at the first cell handed over or, new here, at the cell nearest the point passed to `setStartAndEndPoint`, and returns to that point — the contract `RoutePlannerBase` already offers. From there each step takes the cheapest next cell; it does not try every order.

A single cell has no order to choose and is untouched.

## Two scenes from the tests

Real library output on this branch: same field, same cells, same swaths on both sides, only the route differs. Only the drives between cells are drawn, since the turns inside a cell are the same on both sides. Cell 1 is the given start in both.

The cells are handed over left, right, middle. Today the route passes the middle cell to reach the far one, then drives back for it.

pr8_order

Two rooms joined only at the far end. From the start, the cell behind the wall is 20 m away in a straight line and 210 m to drive. Measuring through the headland graph is what keeps the route from going there first.

pr8_wall

Where the cells already arrive in a good order — the same rooms handed over from the far left — today and this PR drive the same route.

## On real fields

350 fields with OSM boundaries, `ConstHL`, `BoustrophedonDecomp(π/2)`, a 4 m implement, the route driven through `planPath` with Dubins turns. 126 fields come out as a single cell and are untouched. On the other 224 — 5 cells at the median — today's library against this branch:

| over the 224 fields | today | this PR | per field |
|---|---:|---:|---|
| empty travel | 345.3 km | **272.8 km** | 15.5 % less at the median; more on 5 fields |
| path length | 2 035.0 km | **1 964.5 km** | 2.9 % less at the median; longer on 20 fields, by 2.9 % at most |
| driven outside the field | 11 720 m | 11 340 m | more on 21 fields, less on 19 |
| route generation | 71.3 s | 72.8 s | |

**Why the graph distance.** A probe that implements the same rule both ways, on the same 224 fields: with the graph distance it saves 15.3 % of empty travel at the median and does worse than today on 4 fields; with a straight line it saves 12.0 % and does worse on 20. That is the wall scene at field scale.

## Tests

`multicell_route_test.cpp`:

- `visitsCellsInAReasonableOrder`, `doesNotMistakeACellBehindAWallForANearOne` — the two scenes above
- `entersACellFromTheNearerEnd`, `drivesEachCellInOneBlock`
- `startsAtTheCellNearestTheStartPoint`, `connectsBothEndsToTheStartPoint`, `withoutAStartPointTheRouteIsUnchanged`
- `fallsBackToStraightLineWhereTheGraphCannotReach`, `emptyCellsAreSkipped`, `singleCellIsUntouched`
- `everyPatternOrdersItsCells` — the order lives in the base class, so every pattern gets it

`multicell_route_test.py` reaches `setStartAndEndPoint` from Python.


u33549:
The one red job — `g++ 12, swig v4.2.1, shared OFF, Release` — is red on `develop` as well: at `970a1b2`, and on #253 and #254 when they were merged. It fails before anything is compiled:

```
install(EXPORT "Fields2CoverTargets" ...) includes target "Fields2Cover"
which requires target "nlohmann_json" that is not in any export set.
```

and the same for matplot's `cimg` and `nodesoup`.

It became visible in f154d70. Until then the workflow passed `-DBUILD_SHARED_LIBS={{ matrix.shared }}` without the `$`, so CMake was handed a literal string, read it as true, and that row built shared like every other one. With `${{ matrix.shared }}` it builds static for the first time, and a static `Fields2Cover` carries its `PRIVATE` dependencies into the export set — which the FetchContent ones are not part of.

